PR #436, opened 2018-12-10
Fix dockerfile

Herklos:
- Fix dockerfile (Werker CI)


Herklos:
ENTRYPOINT ["python", "start.py", "-ng"]

PR #445, opened 2018-12-16
Coding style improvements

Herklos:
- Fix `is not None` and `is None` conditions
- Improvements of some loops with comprehensive python


Herklos:
It seems that the `is not None` replacement has unexpected effects. I will probably rollback the associated changes.

Herklos:
```
if self.symbol_ticker is not None:
    return True
return False
```

is **different** from : 

`return True if self.symbol_ticker is not None else False`

PR #467, opened 2018-12-19
Improve constants imports

Herklos:
Improve constants imports by moving them in config/__init__.py


GuillaumeDSM:
Looks like CI fails :/

PR #470, opened 2018-12-19
[Docker] Implement persistent configuration

Herklos:
#456 Implement persistent configuration

https://github.com/Drakkar-Software/OctoBot/wiki/With-Docker

### DIFF
--- a/start.py
+++ b/start.py
@@ -84,7 +84,7 @@ def start_octobot(starting_args):
             else:
                 if not tentacles_arch_exists():
                     logger.info("No tentacles found installing default...")
-                    Commands.package_manager(config, ["install", "all"])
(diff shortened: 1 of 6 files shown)


Herklos:
It's done except if the config.json file doesn't exists at $(pwd)/config.json.

Herklos:
Found a temporary solution with `touch config.json`.

start.py
```diff
@@ -84,7 +84,7 @@ def start_octobot(starting_args):
             else:
                 if not tentacles_arch_exists():
                     logger.info("No tentacles found installing default...")
-                    Commands.package_manager(config, ["install", "all"])
+                    Commands.package_manager(config, ["install", "all"], force=True)
```
GuillaumeDSM: very good idea !

Issue #474, opened 2018-12-20
[Exchange] Ensure market data is not 'nan'


Herklos:
- Currently on binance -> min / max price = nan

Herklos:
See https://github.com/ccxt/ccxt/issues/4325

Herklos:
Proposition: 
- Calculate max_price (if nan) = max_cost / max_qty
- Calculate min_price (if nan) = min_cost / min_qty

Herklos:
https://github.com/Drakkar-Software/OctoBot/issues/104

PR #480, opened 2018-12-22
#474 Implement exchange market status  fixer

Herklos:
#474 Implement exchange market status fixer

### DIFF
--- a/tests/unit_tests/trading_tests/test_exchange_market_status_fixer.py
+++ b/tests/unit_tests/trading_tests/test_exchange_market_status_fixer.py
@@ -1,0 +1,244 @@
+#  Drakkar-Software OctoBot
+#  Copyright (c) Drakkar-Software, All rights reserved.
+#
+#  This library is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU Lesser General Public
+#  License as published by the Free Software Foundation; either
+#  version 3.0 of the License, or (at your option) any later version.
+#
+#  This library is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+#  Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser General Public
+#  License along with this library.
+from math import nan
+
+from config import ExchangeConstantsMarketStatusColumns as Ecmsc
+from config import ExchangeConstantsMarketStatusInfoColumns as Ecmsic
+from trading import ExchangeMarketStatusFixer
+
+
+class TestExchangeMarketStatusFixer:
+
+    @staticmethod
+    def _get_limits(amount_min=None, amount_max=None,
+                    price_min=None, price_max=None,
+                    cost_min=None, cost_max=None):
+        return {
+            Ecmsc.LIMITS_AMOUNT.value: {
+                Ecmsc.LIMITS_AMOUNT_MIN.value: amount_min,
+                Ecmsc.LIMITS_AMOUNT_MAX.value: amount_max,
+            },
+            Ecmsc.LIMITS_PRICE.value: {
+                Ecmsc.LIMITS_PRICE_MIN.value: price_min,
+                Ecmsc.LIMITS_PRICE_MAX.value: price_max,
+            },
+            Ecmsc.LIMITS_COST.value: {
+                Ecmsc.LIMITS_COST_MIN.value: cost_min,
+                Ecmsc.LIMITS_COST_MAX.value: cost_max,
+            },
+        }
+
+    @staticmethod
+    def _get_precision(precision_amount=None, precision_cost=None, precision_price=None):
+        return {
+            Ecmsc.PRECISION_AMOUNT.value: precision_amount,
+            Ecmsc.PRECISION_COST.value: precision_cost,
+            Ecmsc.PRECISION_PRICE.value: precision_price,
+        }
+
+    # Global
+    def test_exchange_market_status_fixer_without_precision_cost_and_amount_with_price(self):
+        current_price = 4564.1458
+        ms = {
+            Ecmsc.PRECISION.value: self._get_precision(nan, None, nan),
+            Ecmsc.LIMITS.value: self._get_limits(None, nan, 0.05, 1e4, nan, None)
+        }
+
+        assert ExchangeMarketStatusFixer(ms, price_example=current_price).get_market_status() == {
+            Ecmsc.PRECISION.value: self._get_precision(4, 4, 4),
+            Ecmsc.LIMITS.value: self._get_limits(0.002190990480628382, 219.0990480628382,
+                                                 current_price / 1000, current_price * 1000,
+                                                 0.010000000000000012, 1000000000.0000011)
+        }
+
+        current_price = 1.56e-6
+        ms = {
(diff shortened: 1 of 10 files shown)


Herklos:
Using binance info from ccxt : 
capture

trading/exchanges/exchange_market_status_fixer.py
```diff
@@ -0,0 +1,265 @@
...
+        self.market_status[Ecmsc.LIMITS.value] = {
+            Ecmsc.LIMITS_AMOUNT.value: {
+                Ecmsc.LIMITS_AMOUNT_MIN.value: amount_min,
+                Ecmsc.LIMITS_AMOUNT_MAX.value: amount_max,
+            },
+            Ecmsc.LIMITS_PRICE.value: {
+                Ecmsc.LIMITS_PRICE_MIN.value: price_min,
+                Ecmsc.LIMITS_PRICE_MAX.value: price_max,
+            },
+            Ecmsc.LIMITS_COST.value: {
```
GuillaumeDSM: isn't it supposed to be price*amount ?

tests/unit_tests/trading_tests/test_exchange_market_status_fixer.py
```diff
@@ -0,0 +1,239 @@
...
+
+        current_price = 1.56e-6
+        ms = {
+            Ecmsc.PRECISION.value: self._get_precision(nan, None, nan),
+            Ecmsc.LIMITS.value: self._get_limits(None, nan, 0, 0, nan, None)
+        }
+
+        assert ExchangeMarketStatusFixer(ms, price_example=current_price).get_market_status() == {
+            Ecmsc.PRECISION.value: self._get_precision(8, 8, 8),
+            Ecmsc.LIMITS.value: self._get_limits(513231262.1957913, 345077731.54405594,
```
GuillaumeDSM: 513231262.1957913 is min amount for something with a price of 1.56e-6  ? => if price in BTC, min amount is 800.64 BTC ?

Herklos: Fixed !

trading/exchanges/exchange_market_status_fixer.py
Herklos: Yes !

trading/exchanges/exchange_market_status_fixer.py
```diff
@@ -0,0 +1,270 @@
...
+
+        if amount_log_price >= 0:
+            amount_min = 10 ** (self.LIMIT_AMOUNT_MIN_SUP_ATTENUATION - amount_log_price)
+        else:
+            amount_min = 10 ** -(amount_log_price + self.LIMIT_AMOUNT_MIN_ATTENUATION)
+
+        if amount_log_price >= 0:
+            amount_max = 10 ** (self.LIMIT_AMOUNT_MAX_SUP_ATTENUATION - amount_log_price)
+        else:
+            amount_max = 10 ** (-amount_log_price + self.LIMIT_AMOUNT_MAX_MINUS_3_ATTENUATION)
```
GuillaumeDSM: merge those two if amount_log_price >= 0: conditions for clarity ?

tests/unit_tests/trading_tests/test_exchange_market_status_fixer.py
```diff
@@ -0,0 +1,244 @@
...
+    def test_exchange_market_status_fixer_without_precision_cost_and_amount_with_price(self):
+        current_price = 4564.1458
+        ms = {
+            Ecmsc.PRECISION.value: self._get_precision(nan, None, nan),
+            Ecmsc.LIMITS.value: self._get_limits(None, nan, 0.05, 1e4, nan, None)
+        }
+
+        assert ExchangeMarketStatusFixer(ms, price_example=current_price).get_market_status() == {
+            Ecmsc.PRECISION.value: self._get_precision(4, 4, 4),
+            Ecmsc.LIMITS.value: self._get_limits(0.002190990480628382, 219.0990480628382,
```
GuillaumeDSM: 0.002190990480628382 * 4564.1458 = 10    =>   means that when current_price > 1: we consider 10 to be min total cost (which is usually the case in XXX/USD market) but when price < 1: we consider 0.001 to be min total cost (which is usually the case in XXX/BTC market). 
Since we know the reference market ('baseAsset' in ccxt market status), we should be able to call the appropriate one in _calculate_amount(): 10 if USD, .001 if btc, etc. 
However it's not a perfect solution but unfortunatelly the best I can think of (we would need to add common base markets: BTC, USDT, ETH, XRP, NEO).
Another solution could be to let it as is and consider that everyone is trading vs BTC or USD(T) and that BTC is always the most expensive crypto regarding 1 coin absolute value.

Herklos: I think that each base market can be calculated the same way.

If XXX/USD is 4564 USD --> min = 0.002190990480628382 so actual min is 10 USD
if YYY/BTC is 4564 BTC --> "" so min is 10 BTC
even if ZZZ/ETH is 4564 ETH --> 10 ETH min

I don't think we need to perform different solution for each base market.

trading/exchanges/exchange_market_status_fixer.py
Herklos: How ?

tests/unit_tests/trading_tests/test_exchange_market_status_fixer.py
GuillaumeDSM: "so min is 10 BTC" that's not possible, is it ? min order is 40K USD ?

Herklos: No I took the example for YYY/BTC (not YYY/USD) so it will be 4564BTC for each YYY so minium will be 10BTC so its logic.

I think don't need to calculate different way because it's the same quantity, we just change the base market.

Herklos: The algorithm doesn't need to be changed for each base market. Minimum and maximum will be equals for 1000USD/BTC and 1000ETH/BTC or 1000BTC/USD.

GuillaumeDSM: Absolutely and I think this is where the problem is: it's not the same value for each base asset:
[binance min structure](https://support.binance.com/hc/en-us/articles/115000594711-Trading-Rule). 
ex with min trading value in binance: for the same symbol: 0.001 BTC, 0.01 ETH, 1 BNB, 10 USDT, 1 TUSD. Here with the current system, 

Therefore we need to be carefull when computing minimum amount using only the price as a reference, i think we can't compute it using only the price, we also need base asset: 
formula should be something like 
min_amount=base_asset_min/price_example 
if we take an example, lets say NEO, it would be:
NEO/BTC
NEO/BTC
NEO/BTC
NEO/BTC

Herklos: Yes, but the idea here is not to calculate the real value of the minimum because it's really different in each exchange. But we create a filter, the bot will only try to create orders with relevent quantity, and if this order is on a symbol with a higher minimum the order will be discarded and the bot will take this into account. I really think that it's not a problem because it will not introduce abnormal behavior. This feature will just impact the filter implemented previously.

GuillaumeDSM: Ok for me then

PR #482, opened 2018-12-23
Docker images best practices improvements

Herklos:


GuillaumeDSM:
Need to create logs folder ? looks like docker is not creating it now that it's ignored

Herklos:
Yes exactly this CI is no mounting volumes, I will have a lot at this problem.

Herklos:
Exactly !

PR #575, opened 2019-01-19
Improve UX

GuillaumeDSM:


Herklos:
Can you add some screenshots ?

services/notifier_service_factory.py
```diff
@@ -38,4 +38,7 @@ def get_service_key_to_be_created(self):
 
     @staticmethod
     def get_notifiers_instance(config):
-        return config[CONFIG_CATEGORY_SERVICES][CONFIG_NOTIFIER][CONFIG_SERVICE_INSTANCE]
+        if CONFIG_NOTIFIER in config[CONFIG_CATEGORY_SERVICES] \
```
Herklos: Why do we need to have "notifier" in config services ?

PR #599, opened 2019-01-25
Feature/feedback system

GuillaumeDSM:
image

### DIFF
--- a/tools/external_links_manager.py
+++ b/tools/external_links_manager.py
@@ -15,17 +15,12 @@
 #  License along with this library.
 
 import json
+import requests
 
-from . import api
-from config import PROJECT_NAME, LONG_VERSION, EXTERNAL_LINK_CURRENT_USER_FORM
-from tools import external_links_manager
-
-
-@api.route("/version")
-def version():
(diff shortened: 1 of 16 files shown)


tools/external_links_manager.py
```diff
@@ -0,0 +1,26 @@
...
+#  You should have received a copy of the GNU Lesser General Public
+#  License along with this library.
+
+import json
+import requests
+
+from config import GITHUB_RAW_CONTENT_URL, ASSETS_BRANCH, GITHUB_REPOSITORY, EXTERNAL_LINKS_FILE
+
+
+def get_external_link(link_key):
```
Herklos: I don't really see the interest of creating a new tool for this.

GuillaumeDSM: It's not a specific feature for the web interface and can be reused for any server-side update without updating the client software.
Maybe another name would be more appropriate, something like online resource-manager ?
I'm thinking about possibilities like info messages to broadcast to all users for example

GuillaumeDSM: What do you think of the correction I made ?

Herklos: Perfect 👍

PR #604, opened 2019-01-26
#603 added starting message

GuillaumeDSM:
image


Herklos:
I don't understand why the docker build is failing...

Issue #578, opened 2019-01-19
[Bug][Async] Can't stop OctoBot properly


Herklos:
```
Traceback (most recent call last):
  File "/home/herklos/dev/Drakkars/OctoBot/evaluator/Updaters/global_price_updater.py", line 163, in _refresh_time_frame_data
    await self._refresh_data(time_frame, symbol, notify=notify)
  File "/home/herklos/dev/Drakkars/OctoBot/evaluator/Updaters/global_price_updater.py", line 177, in _refresh_data
    return_list=False))
  File "/home/herklos/dev/Drakkars/OctoBot/trading/exchanges/exchange_dispatcher.py", line 89, in get_symbol_prices
    await self.exchange.get_symbol_prices(symbol=symbol, time_frame=time_frame, limit=limit)
  File "/home/herklos/dev/Drakkars/OctoBot/trading/exchanges/rest_exchanges/rest_exchange.py", line 126, in get_symbol_prices
    candles = await self.client.fetch_ohlcv(symbol, time_frame.value)
  File "/home/herklos/.local/lib/python3.7/site-packages/ccxt/async_support/binance.py", line 503, in fetch_ohlcv
    response = await self.publicGetKlines(self.extend(request, params))
  File "/home/herklos/.local/lib/python3.7/site-packages/ccxt/async_support/binance.py", line 1114, in request
    response = await self.fetch2(path, api, method, params, headers, body)
  File "/home/herklos/.local/lib/python3.7/site-packages/ccxt/async_support/base/exchange.py", line 113, in fetch2
    await self.throttle()
concurrent.futures._base.CancelledError
 2019-01-19 19:05:51,432 ERROR    GlobalPriceUpdater    when refreshing data for time frame TimeFrames.FOUR_HOURS: 
 2019-01-19 19:05:51,432 ERROR    GlobalPriceUpdater   
Traceback (most recent call last):
  File "/usr/lib/python3.7/asyncio/runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "/usr/lib/python3.7/asyncio/base_events.py", line 584, in run_until_complete
    return future.result()
concurrent.futures._base.CancelledError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/herklos/dev/Drakkars/OctoBot/evaluator/Updaters/global_price_updater.py", line 163, in _refresh_time_frame_data
    await self._refresh_data(time_frame, symbol, notify=notify)
[...]

PR #616, opened 2019-01-26
#578 now gracefully stops

GuillaumeDSM:


tools/commands.py
```diff
@@ -14,18 +14,24 @@
...
+from tools.logging.logging_util import get_logger
 from backtesting.collector.data_collector import DataCollector
 from config.config import encrypt
 from tools.tentacle_creator.tentacle_creator import TentacleCreator
 from tools.tentacle_manager.tentacle_manager import TentacleManager
 
 
 class Commands:
+
+    BOT = None
```
Herklos: I think this is not a bad idea, but we need to uniform this with other part of the program.

For example, for interfaces we added a `bot_instance`  in __init__.py and a function `get_bot()`.

 Do you think we can do the same here ?

GuillaumeDSM: Absolutely, much better like this

PR #633, opened 2019-01-29
#631 fixed dusts management round system

GuillaumeDSM:

PR #634, opened 2019-01-29
[Version] 0.3.1

Herklos:
depends on #633


GuillaumeDSM:
then i will remove dev as tentacle branch in #633, i also have a minor quality improvement to add then we merge

PR #637, opened 2019-01-29
Fix travis deployment without testing

Herklos:
- If setup.py build script pass deployment is triggered without waiting for pytest result.

See https://travis-ci.org/Drakkar-Software/OctoBot/jobs/486100232


GuillaumeDSM:
Why not waiting fot pytest results ?

Herklos:
> Why not waiting fot pytest results ?

Here travis will deploy when each matrix job ends. But a new job for setup.py build test was introduced in https://github.com/Drakkar-Software/OctoBot/pull/632. At the end of this job, it trigger the deployment, but it doesn't contains any tests so it shouldn't.

Issue #655, opened 2019-02-07
[Configuration interface] do not display in dev evaluators


GuillaumeDSM:

GuillaumeDSM:
Change applied in OctoBot Tentacles project, see: https://github.com/Drakkar-Software/OctoBot-Tentacles/commit/bb87810abf48cf44ab0d1e034261033b01256ff2

PR #656, opened 2019-02-07
#655 added dev-mode check in evaluators display

GuillaumeDSM:
image

no info => activated

class param
    DEV_MODE = True
=> not displayed

### DIFF
--- a/evaluator/abstract_evaluator.py
+++ b/evaluator/abstract_evaluator.py
@@ -28,6 +28,7 @@ from tools.config_manager import ConfigManager
 class AbstractEvaluator:
     __metaclass__ = ABCMeta
 
+    DEV_MODE = False
     DESCRIPTION = "No description set."
 
     def __init__(self):
@@ -109,6 +110,11 @@ class AbstractEvaluator:
     # Description of the evaluator, used as documentation
     def get_description(cls):
         return cls.DESCRIPTION
+
+    @classmethod
+    # True if evaluator is currently in development (not usable yet), default is False
+    def get_dev_mode(cls):
+        return cls.DEV_MODE
 
     # If this indicator is enabled
     def get_is_enabled(self):
--- a/interfaces/web/models/configuration.py
+++ b/interfaces/web/models/configuration.py
@@ -168,7 +168,7 @@ def _fill_evaluator_config(evaluator_name, activated, eval_type_key,
     description = "description"
     advanced_class_key = "advanced_class"
     klass = get_class_from_string(evaluator_name, AbstractEvaluator, evaluator_type, evaluator_parent_inspection)
-    if klass:
+    if klass and not klass.get_dev_mode():
         detailed_config[eval_type_key][evaluator_name] = {}
         detailed_config[eval_type_key][evaluator_name][EVALUATOR_ACTIVATION] = activated
         detailed_config[eval_type_key][evaluator_name][description] = klass.get_description()


Herklos:
I don't think this is the best way to check if the current tentacle is in dev mode.

This information is provided in the json description.

GuillaumeDSM:
> I don't think this is the best way to check if the current tentacle is in dev mode.
> 
> This information is provided in the json description.

On the scale of a tentacle yes but not on multi evaluators tentacles

Herklos:
> 
> 
> > I don't think this is the best way to check if the current tentacle is in dev mode.
> > This information is provided in the json description.
> 
> On the scale of a tentacle yes but not on multi evaluators tentacles

A multi-evaluator tentacle ?

GuillaumeDSM:
> > I don't think this is the best way to check if the current tentacle is in dev mode.
> > This information is provided in the json description.
> 
> On the scale of a tentacle yes but not on multi evaluators tentacles

Actually, dev_mode is not the right label, it should be 

> > > I don't think this is the best way to check if the current tentacle is in dev mode.
> > > This information is provided in the json description.
> > 
> > 
> > On the scale of a tentacle yes but not on multi evaluators tentacles
> 
> A multi-evaluator tentacle ?

I mean in default tentacles we have files with more than one evaluator => we need a way not to display them all if some are not to be used yet

Herklos:
> 
> 
> > > I don't think this is the best way to check if the current tentacle is in dev mode.
> > > This information is provided in the json description.
> > 
> > 
> > On the scale of a tentacle yes but not on multi evaluators tentacles
> 
> Actually, dev_mode is not the right label, it should be
> 
> > > > I don't think this is the best way to check if the current tentacle is in dev mode.
> > > > This information is provided in the json description.
> > > 
> > > 
> > > On the scale of a tentacle yes but not on multi evaluators tentacles
> > 
> > 
> > A multi-evaluator tentacle ?
> 
> I mean in default tentacles we have files with more than one evaluator => we need a way not to display them all if some are not to be used yet

Hum ok I see, but I think this is not the best solution to duplicate this option (DEV_MODE) because we have multiple evaluator in some files. I think we have to choose between : 
- The description for the whole file (same for each evaluator in file)
- To move evaluator in development in a different file with a new description (and eventually move it back to the original file)
- Don't enable multi-evaluator files.

GuillaumeDSM:
> > > > I don't think this is the best way to check if the current tentacle is in dev mode.
> > > > This information is provided in the json description.
> > > 
> > > 
> > > On the scale of a tentacle yes but not on multi evaluators tentacles
> > 
> > 
> > Actually, dev_mode is not the right label, it should be
> > > > > I don't think this is the best way to check if the current tentacle is in dev mode.
> > > > > This information is provided in the json description.
> > > > 
> > > > 
> > > > On the scale of a tentacle yes but not on multi evaluators tentacles
> > > 
> > > 
> > > A multi-evaluator tentacle ?
> > 
> > 
> > I mean in default tentacles we have files with more than one evaluator => we need a way not to display them all if some are not to be used yet
> 
> Hum ok I see, but I think this is not the best solution to duplicate this option (DEV_MODE) because we have multiple evaluator in some files. I think we have to choose between :
> 
> * The description for the whole file (same for each evaluator in file)
> * To move evaluator in development in a different file with a new description (and eventually move it back to the original file)
> * Don't enable multi-evaluator files.

I think moving them to a different file is the best solution.
A description for the whole file is very unflexible, same with not allowing multi evaluator files.

GuillaumeDSM:
See https://github.com/Drakkar-Software/OctoBot-Tentacles/commits/dev

Herklos:
Perfect ! 👍

PR #487, opened 2018-12-28
Refactor Trade classe - Python 3.7

Herklos:
Issue #486

PR #660, opened 2019-02-09
[Python3.7] Refactor Order and Trade classes

### LINKED ISSUE #486: [Python3.7] Refactor Order and Trade classes
After #485 merge.

Refactor with :
- data_class (https://www.python.org/dev/peps/pep-0557/#id7)
- @property

Herklos:
This PR concerns Order class.

Done for Trade class in https://github.com/Drakkar-Software/OctoBot/pull/487.

resolves #486

### DIFF
--- a/trading/trader/order.py
+++ b/trading/trader/order.py
@@ -18,54 +18,62 @@ import time
 import math
 from abc import *
 from asyncio import Lock
+from dataclasses import dataclass, field
+from typing import List, Any, Dict, Union
+
 from tools.logging.logging_util import get_logger
 
 from tools.symbol_util import split_symbol
 from config import TradeOrderSide, OrderStatus, TraderOrderType, SIMULATOR_LAST_PRICES_TO_CHECK, \
     ExchangeConstantsTickersColumns as eC, FeePropertyColumns, ExchangeConstantsMarketPropertyColumns
+from trading.exchanges.exchange_dispatcher import ExchangeDispatcher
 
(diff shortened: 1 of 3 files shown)


trading/trader/order.py
```diff
@@ -18,54 +18,60 @@
...
+    is_simulated: bool = field(init=False, repr=False)
+    side: TradeOrderSide = None
+    symbol: str = None
+    origin_price: float = 0
+    origin_stop_price: float = 0
+    origin_quantity: float = 0
+    market_total_fees: float = 0
+    filled_quantity: float = 0
+    filled_price: float = 0
+    fee: float = 0
```
GuillaumeDSM: Dict[str, Union[str, float]]

trading/trader/order.py
```diff
@@ -18,54 +18,60 @@
...
+    last_prices: List[Any] = None
+    created_last_price: float = 0
+    order_profitability: float = 0
+    linked_to: Any = None
+    is_from_this_octobot: bool = True
+    linked_portfolio: Any = None
+    order_notifier: Any = None
+    linked_orders: List[Any] = List
+    taker_or_maker: ExchangeConstantsMarketPropertyColumns = None
+    lock: Lock = Lock()
```
GuillaumeDSM: Looks like we can't initialize an asyncio.Lock in a dataclass that way (from CI)

trading/trader/order.py
Herklos: Yes !

PR #668, opened 2019-02-14
Remove performance analyser

Herklos:
Replaced by `docker stats` when using an OctoBot container.


GuillaumeDSM:
I agree to use docker stats but why removing the possibility to measure perfs on non-docker platforms ? To me it can be useful.

PR #695, opened 2019-02-23
prepare for staggered orders strategy

GuillaumeDSM:
see https://github.com/Drakkar-Software/OctoBot-Tentacles/issues/28


GuillaumeDSM:
problem: max_digits not take into account (seen on rdn/eth [binance])

PR #718, opened 2019-03-10
#717 now remove orderes that are absent from exchange but still in or…

GuillaumeDSM:
depends on #716


GuillaumeDSM:
moved to #721

PR #732, opened 2019-03-16
improved python version check

GuillaumeDSM:
image

### DIFF
--- a/tools/__init__.py
+++ b/tools/__init__.py
@@ -28,12 +28,11 @@ if not current_version >= MIN_PYTHON_VERSION:
           + "." + str(current_version[1]) + "\n"
           + "You can download Python last versions on: https://www.python.org/downloads/", file=sys.stderr)
     sys.exit(-1)
+else:
+    # binary tentacle importation
+    sys.path.append(os.path.dirname(sys.executable))
 
-# binary tentacle importation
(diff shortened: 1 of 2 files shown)


tools/__init__.py
```diff
@@ -14,29 +14,30 @@
 #  You should have received a copy of the GNU Lesser General Public
 #  License along with this library.
 
-import logging
+from __future__ import print_function
```
Werdor: Unused ?

tools/__init__.py
```diff
@@ -14,29 +14,30 @@
...
+          + "." + str(MIN_PYTHON_VERSION[1]) + " current Python version is " + str(current_version[0])
+          + "." + str(current_version[1]) + "\n"
+          + "You can download Python last versions on: https://www.python.org/downloads/", file=sys.stderr)
     sys.exit(-1)
 
 # binary tentacle importation
 sys.path.append(os.path.dirname(sys.executable))
 
 # if compatible version, can proceed with imports
+if True:
```
Werdor: Why do you need `if True` ?
the comment is about `if compatible version`.

tools/__init__.py
GuillaumeDSM: used to locally redefine print (to print in sys.err)

PR #733, opened 2019-03-17
WIP Feature/trader persistence

GuillaumeDSM:
Are now kept between instances (if symbol config did not change):
* simulator & real traders trade history
* initial portfolio (for initial/no trade portfolio profitability & after trade profitability)
* watched symbols profitability 

bot started after: 03:34:24
image

reset history from interface
image

Issue #731, opened 2019-03-16
[Trader] Allow to start from the previous execution last move


GuillaumeDSM:
A setting in config.json (trading section), checkbox in web interface, unticked by default

If activated:
In a file:
* store very first startup portfolio that will be kept as a reference 
* store very first startup portfolio value and watched markets origin values
* store reference market (for consistency)

When starting the bot, if this option is selected, use this file to setup the simulator initiatial portfolio and keep profitability between instances
* Use log files to restore last portfolio

To reset this history:
* change symbol configuration
* reset button on web interface
* start the bot with --reset-history parameter
* delete file

PKBO:
+ add / keep option to clear this history, when strategy changes

PR #735, opened 2019-03-17
FOR REVIEW #731 added history handle for profitability and tradres for real and …

GuillaumeDSM:
for review only
extract of 9ba3a8f2ffa36f672bfbb54cf30b38c6e3f92b4b from #733

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -122,3 +122,5 @@ config\.json
 *.ods
 temp_config\.json
 log
+
+config/trading_state_history\.json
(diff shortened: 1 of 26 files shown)


.gitignore
```diff
@@ -122,3 +122,5 @@ config\.json
 *.ods
 temp_config\.json
 log
+
+config/trading_state_history\.json
```
Werdor: This is not possible (this path) with the current docker volume management.

Werdor: A good feature to solve this problem can be a folder for config.json and this file.

GuillaumeDSM:
moved to #754

PR #742, opened 2019-03-19
Cancel orders from web interface

GuillaumeDSM:
image
image
image


interfaces/trading_util.py
```diff
@@ -114,13 +114,21 @@ def get_open_orders():
 
     for trader in traders:
         if trader.get_simulate():
-            simulated_open_orders.append(trader.get_open_orders())
+            simulated_open_orders += trader.get_open_orders()
```
Werdor: Why not append ?

GuillaumeDSM: [].append([xyz]) = [[xyz]]   which makes no sense here, we want [xyz]

PR #745, opened 2019-03-20
Fix trader unused OM

Werdor:


GuillaumeDSM:
We also  need this in octobot.py:
```python
async def start_tasks(self, run_in_new_thread=False):
...
            if trader.enable:
                task_list.append(trader.start_order_manager())
...
            if trader_simulator.enable:
                task_list.append(trader_simulator.start_order_manager())
..
```

PR #757, opened 2019-03-28
#756 ensured call_order_update_callback is always called

GuillaumeDSM:


trading/trader/orders_manager.py
```diff
@@ -196,17 +196,23 @@ def set_order_refresh_time(self, seconds):
...
+                    # update symbol prices from exchange
+                    if odr.get_order_symbol() in self.last_symbol_prices:
+                        odr.set_last_prices(self.last_symbol_prices[odr.get_order_symbol()])
+
+                    if odr in self.order_list:
+                        order_filled = await self._update_order_status(odr, failed_order_updates, simulated_time)
+                if order_filled:
+                    await self.trader.call_order_update_callback(order)
+            except Exception as e:
+                # ensure always call fill callback
```
Werdor: Why not : 
```
finally:
    if order_filled:
        await self.trader.call_order_update_callback(order)
```

GuillaumeDSM: because finally won't be called after raise e

PR #758, opened 2019-03-30
Refactor octobot py

Werdor:
Refactor octobot.py splitted in 4 new files : 
- initializer.py
- exchange_factory.py
- evaluator_factory.py
- task_manager.py


core/initializer.py
```diff
@@ -0,0 +1,93 @@
...
+from evaluator.Util import AdvancedManager
+from evaluator.evaluator_creator import EvaluatorCreator
+from services import ServiceCreator
+from tools import get_logger
+from tools.notifications import Notification
+from tools.performance_analyser import PerformanceAnalyser
+from tools.time_frame_manager import TimeFrameManager
+
+
+class Initializer:
```
GuillaumeDSM: I think it would be easier to read with _ prefixes for protected methods

GuillaumeDSM: Good idea to wrap initalizations in one file :)

setup.py
```diff
@@ -1,4 +1,4 @@
-#  Drakkar-Software OctoBot-Launcher
```
GuillaumeDSM: 👍

core/initializer.py
Werdor: Yes good idea ;)

PR #766, opened 2019-04-03
Added telemetry

GuillaumeDSM:
Using https://github.com/Drakkar-Software/Metrics-Server

image


core/octobot.py
```diff
@@ -47,6 +47,9 @@ def __init__(self, config, ignore_config=False, reset_trading_history=False):
             BOT_TOOLS_RECORDER: None,
         }
 
+        # metrics if enabled
+        self.metrics_handler = None
```
Werdor: Why not in Initializer ?

GuillaumeDSM: To me it seems more legit to store it in octobot

tools/metrics_manager.py
```diff
@@ -0,0 +1,116 @@
...
+        first_evaluator = next(iter(self.octobot.get_symbols_tasks_manager().values())).get_evaluator()
+        evaluators = copy.copy(first_evaluator.get_social_eval_list())
+        evaluators += first_evaluator.get_ta_eval_list()
+        evaluators += first_evaluator.get_real_time_eval_list()
+        for evaluator in evaluators:
+            config_eval.append(evaluator.get_name())
+        return config_eval
+
+    def _init_bot_id(self):
+        answer = requests.get(f"{METRICS_URL}{METRICS_ROUTE_GEN_BOT_ID}", headers=self._headers)
```
Werdor: There is no way to use requests or equivalent with async ?

GuillaumeDSM: There should be, I need to look into it

core/octobot.py
Werdor: Ok :+1:

PR #767, opened 2019-04-04
Added utility methods

GuillaumeDSM:
used by: https://github.com/Drakkar-Software/OctoBot-Tentacles/pull/48


config/__init__.py
```diff
@@ -265,7 +265,7 @@
 TENTACLE_PACKAGE_DESCRIPTION = "package_description"
 EVALUATOR_CONFIG_FOLDER = "config"
 TENTACLES_TRADING_MODE_PATH = "Mode"
-TENTACLES_DEFAULT_BRANCH = "master"
+TENTACLES_DEFAULT_BRANCH = "dev"
```
Werdor: That was for your tests or we need to keep this before releasing ?

trading/trader/modes/abstract_mode_creator.py
```diff
@@ -245,6 +252,19 @@ def check_and_adapt_order_details_if_necessary(quantity, price, symbol_market, f
...
             return []
 
+    @staticmethod
+    async def get_holdings_ratio(trader, portfolio, currency):
+        pf_copy = deepcopy(portfolio.get_portfolio())
+        pf_value = await trader.get_trades_manager().update_portfolio_current_value(pf_copy)
+        currency_holdings = Portfolio.get_currency_from_given_portfolio(pf_copy, currency,
+                                                                        portfolio_type=Portfolio.TOTAL)
+        currency_value = await trader.get_trades_manager().evaluate_value(currency, currency_holdings)
+        return currency_value / pf_value
```
Werdor: You don't need to check if pf_value > 0 ?

config/__init__.py
GuillaumeDSM: We will need to keep this till the next release because DailyTradingMode now uses get_number_of_traded_assets and get_holdings_ratio that exists only in the dev branch of OctoBot => tentacles have to stay in dev otherwise regular users will not be able to use DailyTradingMode

Werdor: Yes ok !

PR #774, opened 2019-04-06
fixed get_traded_pairs limit case

GuillaumeDSM:


trading/exchanges/exchange_manager.py
```diff
@@ -173,6 +174,7 @@ def _load_config_symbols_and_time_frames(self):
 
     # SYMBOLS
     def _set_config_traded_pairs(self):
+        self.cryptocurrencies_traded_pairs = {}
```
Werdor: Already in __init__, required ?

GuillaumeDSM: Since it's a set, i think its better to reset the values, now its not required

trading/exchanges/exchange_manager.py
```diff
@@ -103,6 +103,7 @@ def need_user_stream(self):
         # if simulated : create exchange simulator instance
         else:
             self.exchange = ExchangeSimulator(self.config, self.exchange_type, self)
+            self._set_config_traded_pairs()
```
Werdor: 👍

PR #775, opened 2019-04-06
Fix pytest warning summary

Werdor:


GuillaumeDSM:
looks like travis doesn't want to print warnings when duration is enabled

PR #803, opened 2019-04-18
added reference market getter from trader

GuillaumeDSM:
required by https://github.com/Drakkar-Software/OctoBot-Tentacles/pull/56

### DIFF
--- a/trading/trader/trader.py
+++ b/trading/trader/trader.py
@@ -363,8 +363,7 @@ class Trader(Initializable):
             # debug purpose
             profitability_str = PrettyPrinter.portfolio_profitability_pretty_print(profitability,
                                                                                    profitability_percent,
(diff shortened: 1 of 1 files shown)


trading/trader/trader.py
```diff
@@ -363,8 +363,7 @@ def create_order_instance(self, order_type, symbol, current_price, quantity,
             # debug purpose
             profitability_str = PrettyPrinter.portfolio_profitability_pretty_print(profitability,
                                                                                    profitability_percent,
-                                                                                   self.get_trades_manager().
```
Herklos: The `trades_manager` parameter was only used to get the `reference_market` ?

GuillaumeDSM: self.get_trades_manager() is also used for self.get_trades_manager().get_profitability(), etc

PR #808, opened 2019-04-18
add CodeOwners

Herklos:


.github/CODEOWNERS
```diff
@@ -0,0 +1,3 @@
+# These owners will be the default owners for everything in the repo.
+*       @GuillaumeDSM
+*       @herklos
```
GuillaumeDSM: @Herklos  instead of @herklos  ?

Herklos: Yes !

PR #822, opened 2019-04-26
trading pair and timeframe error message

GuillaumeDSM:
image


trading/exchanges/exchange_manager.py
```diff
@@ -191,6 +193,10 @@ def _set_config_traded_pairs(self):
                     self.cryptocurrencies_traded_pairs[cryptocurrency] += self._add_tradable_symbols(cryptocurrency)
 
             # add to global traded pairs
+            currency_traded_pairs = self.cryptocurrencies_traded_pairs[cryptocurrency]
+            if not currency_traded_pairs:
```
Herklos: Why not only `if not self.cryptocurrencies_traded_pairs[cryptocurrency]` ?

PR #836, opened 2019-04-27
Update contributing style

Herklos:


CONTRIBUTING.md
```diff
@@ -3,4 +3,10 @@
...
     Create your PR against the develop branch, not master.
-    New features need to contain unit tests and must be PEP8 conformant (max-line-length = 100).
\ No newline at end of file
+    New features need to contain unit tests and must be PEP8 conformant (max-line-length = 100).
+    
+# OctoBot additional coding style
+
+- Use local variable only if it improve code clarity
+- Always use comprehension list
+- Use `try ... except` instead of `if X is None` when X is rarely None
```
GuillaumeDSM: Use `try ... except` instead of `if` statetements when if is 99% True

Herklos: Yes that's better.

PR #838, opened 2019-04-27
#837 fixed creator crash

GuillaumeDSM:


evaluator/abstract_evaluator.py
```diff
@@ -66,6 +66,12 @@ def get_config_file_name(cls, config_evaluator_type=None):
         return f"{TENTACLES_PATH}/{TENTACLES_EVALUATOR_PATH}/{config_evaluator_type}/{EVALUATOR_CONFIG_FOLDER}" \
             f"/{cls.get_name() + CONFIG_FILE_EXT}"
 
+    @classmethod
+    def get_config_file_error_message(cls, error):
```
Herklos: Where this method is used ?

Herklos: Oh ok in https://github.com/Drakkar-Software/OctoBot-Tentacles/pull/60

Herklos: Why this is not used at all evaluator config load ?

GuillaumeDSM: Because no one coded it in evaluator config load yet :)

PR #871, opened 2019-05-07
#869 created Tentacle class and tentacles_management package

### LINKED ISSUE #869: [Factorize] new "Tentacle" abstract class
Maybe these two function can be inherited from a new "Tentacle" abstract class ?

_Originally posted by @Herklos in https://github.com/Drakkar-Software/OctoBot/pull/856_

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Tentacles/pull/66 for tests

closes #869


evaluator/Util/__init__.py
```diff
@@ -15,5 +15,4 @@
 #  License along with this library.
 
 from .abstract_util import *
-from .advanced_manager import *
```
Herklos: Why ?

Issue #865, opened 2019-05-03
[Exchange config] simplify exchange token config


GuillaumeDSM:
idea:
allow to start with non-encrypted exchange token (auto encrypt and update config.json at 1st read)

PR #872, opened 2019-05-07
#865 now auto encrypts exchange config if necessary and validates con…

GuillaumeDSM:
…fig with schema when any save

image


tools/config_manager.py
```diff
@@ -81,6 +82,30 @@ def validate_config_file(config=None, schema_file=CONFIG_FILE_SCHEMA):
...
 
+    @staticmethod
+    def config_health_check(config):
+        # ensure api key encryption
+        should_replace_config = False
+        if CONFIG_EXCHANGES in config:
+            for exchange, exchange_config in config[CONFIG_EXCHANGES].items():
+                for key in CONFIG_EXCHANGE_ENCRYPTED_VALUES:
+                    try:
+                        if not ConfigManager._handle_encrypted_value(key, exchange_config, verbose=True):
```
Herklos: should_replace_config &= not ConfigManager._handle_encrypted_value(key, exchange_config, verbose=True)

GuillaumeDSM: It would be 
```python
should_replace_config = (not ConfigManager._handle_encrypted_value(key, exchange_config, verbose=True)) and should_replace_config 
```
because in the other way around, when should_replace_config is False, python would not evaluate the second member of the & => therefore not encrypt other key

=> I think it's easier to understand, this way than in one line in this case, what is you opinion ?

PR #906, opened 2019-05-25
added recently closed orders handling

GuillaumeDSM:
required by staggered orders and probably many future strategies


GuillaumeDSM:
> Hum.. Seems OK but I don't see any use (except for staggered maybe)

We at least need it for staggered orders at the moment and we will need it for any other strategy using an order callback as well as an other source of notification (for sync purposes)

Herklos:
OK I think we can merge it for now and create an issue to improve this feature after 0.4.

PR #952, opened 2019-07-21
fix kraken orders

GuillaumeDSM:
Not a fan of the fix lines 236/237 but since Kraken doesn't send the order status at creation and immediatly after, I see no other option to keep complient with the bot requirements

### DIFF
--- a/trading/exchanges/rest_exchanges/rest_exchange.py
+++ b/trading/exchanges/rest_exchanges/rest_exchange.py
@@ -264,7 +264,8 @@ class RESTExchange(AbstractExchange, Initializable):
     def _ensure_order_details_completeness(order, order_required_fields=None):
         if order_required_fields is None:
             order_required_fields = [ecoc.ID.value, ecoc.TIMESTAMP.value, ecoc.SYMBOL.value, ecoc.TYPE.value,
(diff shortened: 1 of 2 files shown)


trading/exchanges/rest_exchanges/rest_exchange.py
```diff
@@ -225,10 +225,17 @@ def get_client(self):
...
+            # Missing order fields after retry handling
+
+            # on some exchange, market orders are not not including price, add it manually to ensure uniformity
             if created_order[ecoc.PRICE.value] is None and price is not None:
                 created_order[ecoc.PRICE.value] = price
 
+            # on some exchange (ex: Kraken), orders are not not including status, add it manually to ensure uniformity
+            # since order status is required and this order just got created, consider it open
+            if created_order[ecoc.STATUS.value] is None:
+                created_order[ecoc.STATUS.value] = OrderStatus.OPEN.value
```
Herklos: It's not already set by order.py#116 ?

GuillaumeDSM: Not here, exchange created orders are parsed in trader.py#parse_exchange_order_to_order_instance()

PR #954, opened 2019-07-23
[Backtesting] Improve recent trades generation

Herklos:
- Added new backtesting trades generation method (only close prices)
- New option in backtesting config :
```
"backtesting": {
        "files": [],
        "trades_generation": "close_only"
 }
```

The default value of trades_generation is "average" (the same as before).

closes #955

PR #956, opened 2019-07-26
[Backtesting] Fixed backtesting recent trades

GuillaumeDSM:
Idea with recent trades, based on https://github.com/Drakkar-Software/OctoBot/pull/954 (backtesting only): 
* Added *candle_range* parameter (default value False) to exchange.get_recent_trades that is used by exchange simulator
  * When True: creates trades using the whole current candle from high to low and its timestamp
  * When False: creates trades using the close value of the previous candle only but the timestamp of the current candle

current candle: the candle directly following the previous candle in the smallest used timeframe (aka candle in construction in real situations)
previous candle: the last candle beeing pushed into evaluators (aka last complete candle)

* When creating an order, set the *candle_range* param to False and this will return recent trades using the close of the previous candle (closest possible to real life)
* When checking orders fill status, leave *candle_range* to its default value (False), this will return orders based on the whole candle, as if the time passed and this candle got constructed during backtesting just before being pushed into evaluators in the next round

### DIFF
--- a/trading/exchanges/abstract_exchange.py
+++ b/trading/exchanges/abstract_exchange.py
@@ -56,7 +56,7 @@ class AbstractExchange:
         pass
 
     @abstractmethod
-    async def get_recent_trades(self, symbol, limit=50):
+    async def get_recent_trades(self, symbol, limit=50, candle_range=True):
         pass
 
     @abstractmethod
--- a/trading/exchanges/exchange_dispatcher.py
+++ b/trading/exchanges/exchange_dispatcher.py
@@ -173,14 +173,14 @@ class ExchangeDispatcher(AbstractExchange):
 
         return symbol_data.get_symbol_order_book()
 
-    async def get_recent_trades(self, symbol, limit=50):
+    async def get_recent_trades(self, symbol, limit=50, candle_range=True):
         symbol_data = self.get_symbol_data(symbol)
 
         if not self._web_socket_available() or not symbol_data.recent_trades_are_initialized():
             if not self._web_socket_available() or \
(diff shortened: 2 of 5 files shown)


Herklos:
Why you said 'based on #954' and you didn't use this PR anywhere ?

GuillaumeDSM:
Because it's base on ideas of 954, however it doesn't depend on it

trading/exchanges/abstract_exchange.py
```diff
@@ -56,7 +56,7 @@ def get_exchange_manager(self):
         pass
 
     @abstractmethod
-    async def get_recent_trades(self, symbol, limit=50):
+    async def get_recent_trades(self, symbol, limit=50, candle_range=True):
         pass
```
Herklos: This modification should be replaced by using #954

GuillaumeDSM: Why ? It's not the same purpose: one is a config value, the other depends on method call context

Herklos: There is no reason to impact abstract_exchange for a backtesting feature.

GuillaumeDSM: There the reason is inheritence.
We need to find a way to differenciate the results of get_recent_trades in order creation and order fill check, this can be done by a param or by calling a different method. Both affect abstract_exchange. 
Now if you have a solution that is only touching exchange simulator, let's go for it but I didn't find any.

Herklos: I agree to your feature proposal but it's a mistake to add a param to an abstract method for a child implementation purpose.

PR #961, opened 2019-08-09
#960 compute backtesting profitability on the last candle

GuillaumeDSM:
fix regression from update after Drakkar-Software/OctoBot#958


Herklos:
Mac worked, we can say 👍

GuillaumeDSM:
Looks like there is a problem with numpy install on linux

GuillaumeDSM:
> Looks like there is a problem with numpy install on linux

fixed with https://github.com/Drakkar-Software/OctoBot/pull/965

PR #1009, opened 2019-12-23
service migrations

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Commons/pull/8


Herklos:
I really doesn't like the initialization way of dispatchers and service in OctoBot.

Herklos:
First because I think that "dispatchers" are an internal feature of OctoBot-Services.
Moreover I would prefer a standalone initializer like the other module "trading_factory" and "evaluator_factory".

octobot/initializer.py
```diff
@@ -31,6 +32,24 @@ def __init__(self, octobot):
     async def create(self):
         # initialize tools
         self.__init_metrics()
+        await self._create_services()
+
+    async def _create_services(self):
```
Herklos: This should be in an "service_factory" or "service_initializer" file.

octobot/evaluator_factory.py
```diff
@@ -50,6 +54,10 @@ def __init__(self, octobot):
                                              exchange_configuration.time_frames)
         await init_evaluator_chan_logger()
 
+    def _create_dispatchers(self):
```
Herklos: This file is reserved for OctoBot-Evaluator API

octobot/octobot.py
```diff
@@ -71,6 +71,7 @@ def __init__(self, config, ignore_config=False, reset_trading_history=False):
         await self.evaluator_factory.initialize()
         await self.exchange_factory.create()
         await self.evaluator_factory.create()
+        await self.task_manager.start_evaluation_util_tasks()
```
Herklos: This line should call the entrypoint of the OctoBot-Service file in OctoBot.

octobot/task_manager.py
```diff
@@ -75,6 +76,10 @@ def __init_uv_loop(self):
         #     self.current_loop_thread = threading.current_thread()
         #     await self.tools_task_group
 
+    async def start_evaluation_util_tasks(self):
```
Herklos: A method like this should be in the OctoBot-Service file

Herklos:
I also think that we should have merge the first commit and create a new pull request.

Herklos:
I thought about it and I think OctoBot-Services is a "low level" module as OctoBot-Websockets is. So it shouldn't appear in OctoBot requirements.txt and shouldn't have an entrypoint in the main module.

- The code you proposed as an entrypoint of OctoBot-Services should be put in your local demo.py to initialize the available services for test purposes.

- OctoBot-Interfaces and OctoBot-Notifications will have their entrypoint in OctoBot main module. They will import OctoBot-Services as a requirement as OctoBot-Trading do for OctoBot-Websockets.

- If we take the example of the web service. A tentacle web service that runs a web server ONLY (Flask for example) should exists in the OctoBot-Services module. An the OctoBot-Interfaces will have the web interface that requires the OctoBot-Services web server tentacle to be started. 

- In the main module (OctoBot) we'll only ask to start the available interfaces in the OctoBot-Interfaces entrypoint. In the OctoBot-Interfaces API we'll create the tentacle interface instance that requires the service to be started. So at this moment and ONLY at this moment (the same process will have to exist for OctoBot-Notifications), a service is started.

GuillaumeDSM:
> First because I think that "dispatchers" are an internal feature of OctoBot-Services.
> Moreover I would prefer a standalone initializer like the other module "trading_factory" and "evaluator_factory".

absolutely, still in WIP on dispatchers, I will move it when they will be migrated to channel arch

GuillaumeDSM:
> I thought about it and I think OctoBot-Services is a "low level" module as OctoBot-Websockets is. So it shouldn't appear in OctoBot requirements.txt and shouldn't have an entrypoint in the main module.
> 
> * The code you proposed as an entrypoint of OctoBot-Services should be put in your local demo.py to initialize the available services for test purposes.
> * OctoBot-Interfaces and OctoBot-Notifications will have their entrypoint in OctoBot main module. They will import OctoBot-Services as a requirement as OctoBot-Trading do for OctoBot-Websockets.
> * If we take the example of the web service. A tentacle web service that runs a web server ONLY (Flask for example) should exists in the OctoBot-Services module. An the OctoBot-Interfaces will have the web interface that requires the OctoBot-Services web server tentacle to be started.
> * In the main module (OctoBot) we'll only ask to start the available interfaces in the OctoBot-Interfaces entrypoint. In the OctoBot-Interfaces API we'll create the tentacle interface instance that requires the service to be started. So at this moment and ONLY at this moment (the same process will have to exist for OctoBot-Notifications), a service is started.

100% agree regarding interfaces and notifications, it will be much better this way, time to refactor :)
The 3rd thing to consider is the evaluator part of services (dispatchers that will turn into channels): they will need to bind to the services instances, I will think about how to handle that properly, an abstract class should do the job.

octobot/__init__.py
```diff
@@ -26,6 +26,12 @@ def __init__(bot, config):
     global_config = config
 
 
+# TODO: find a better way to keep track of the bot instance in octobot module
```
Herklos: :+1:

start.py
```diff
@@ -155,7 +164,7 @@ def start_octobot(starting_args):
             else:
                 if not tentacles_arch_exists():
                     logger.info("No tentacles found. Installing default tentacles ...")
-                    package_manager(config, ["install", "all"], force=True)
+                    # package_manager(config, ["install", "all"], force=True)
```
Herklos: We should think about the package manager migration

start.py
```diff
@@ -104,6 +105,14 @@ def _log_terms_if_unaccepted(config, logger):
         logger.info("Disclaimer accepted by user.")
 
 
+def _disable_interface_from_param(interface_identifier, param_value, logger):
```
Herklos: Nice ! :+1:

octobot/initializer.py
```diff
@@ -28,9 +32,53 @@ def __init__(self, octobot):
         # Logger
         self.logger = get_logger(self.__class__.__name__)
 
+        self.interface_list = []
+        self.notifier_list = []
+
     async def create(self):
         # initialize tools
         self.__init_metrics()
+        await self._create_interfaces()
```
Herklos: I would prefer to have this two create methods in an services dedicated initializer file like `services_initializer.py`. What do you think about it ?

octobot/exchange_factory.py
```diff
@@ -58,6 +59,7 @@ def __init__(self, octobot, ignore_config=False):
                                                            backtesting_files=[os.getenv('BACKTESTING_FILE')])
                     await exchange_factory.create()
                     await init_exchange_chan_logger(exchange_factory.exchange_name)
+                    self.exchange_manager_list.append(exchange_factory.exchange_manager)
```
Herklos: Where do you use this attribute ?

octobot/initializer.py
GuillaumeDSM: In would rather call it ```interface_initializer.py``` service is an internal/tech term we use but I think we should refer to it as interfaces that includes the web/bots interfaces but also connexions to non exchanges websites. To me services does not imply all this, it's too generic.

Agree of a specific file

octobot/exchange_factory.py
GuillaumeDSM: In interfaces, it's the entrypoint to retrieve trading data (using exchange_managers as keys in subsequent API calls)

Herklos: They doesn't use trading API ?

GuillaumeDSM: We could but we need a way to differenciate exchange managers from a regular bot session from the ones of a backtesting bot session within a regular session => maybe we can add a method based on simulated exchanges and not simulated ones to get all instanciated exchange managers from the trading API. What do you think ?
To me it also makes sense that interfaces can access their octobot directly (required also for config and commands like stop/restart etc)

Herklos: Hum ok I see. I'm not really fond of a list of exchange manager outside of OctoBot-Trading and different list of exchange manager based on if they are simulated, in backtesting or something else.

I think your main problem here is that you can't request the trading API for multiple exchange_managers with the same exchange name ? Like if you have an exchange_manager running on binance and another that backtest on binance at the same time.

To this issue, we can imagine to keep using OctoBot-Trading API to request exchange_manager instances but with something like an ID. For example, when you start a backtesting you'll create a new instance of exchange_manager with a unique ID. This exchange_manager instance can be requested with the exchange name and the ID. 

What do you think about having a unique generated ID for each exchange manager that can be used to request the instance from OctoBot-Trading API ?

Herklos: Proposal in https://github.com/Drakkar-Software/OctoBot-Trading/pull/54

GuillaumeDSM: That's a good idea and is solving almost all the problem. The only thing remaining is the way to identify which exchange managers are related to which "bot". I mean in backtesting form the web interface for example, how to get the exchange managers for the current backtesting session (without the ones from the normal "parent" session).

A way to keep track of these backtesting exchange managers would be to keep their ID somewhere, what do you think about storing them in the associated ```Backtesting``` instance ? Then they would be available through the OctoBot-Backtesting API

octobot/evaluator_factory.py
```diff
@@ -50,6 +53,19 @@ def __init__(self, octobot):
                                              exchange_configuration.symbols,
                                              exchange_configuration.time_frames)
         await init_evaluator_chan_logger()
+        # Start service feeds now that evaluators registered their feed requirements
+        await self._start_service_feeds()
+
+    def _initialize_service_feeds(self):
```
Herklos: I don't see any evaluator_factory methods or attributes used by service methods. So I don't think It's a good idea to put services initialization here. We should initialize it in a different file.

octobot/exchange_factory.py
Herklos: Is it still required ? With the exchange ID implementation I don't think so.

octobot/interface_factory.pxd
```diff
@@ -0,0 +1,27 @@
...
+#
+#  You should have received a copy of the GNU Lesser General Public
+#  License along with this library.
+
+
+cdef class InterfaceFactory:
+    cdef public object octobot
+    cdef object logger
+
+    cdef public list interface_list
```
Herklos: Do you prefer `interface_list` or `interfaces` ?

start.py
Herklos: Will be done in the OctoBot-Tentacles-Manager next major version.

octobot/exchange_factory.py
GuillaumeDSM: It's used to get the exchange managers from a given bot in interfaces, we can have a list of exchange ids instead but we need a way to identifie the exchange managers from the current bot (even in backtesting), exchange ids are better in my opinion.

octobot/interface_factory.pxd
GuillaumeDSM: I used interface_list to be consistent with the rest but interfaces is better in my opinion (no need for type in variable name), we can switch if you aggree

octobot/evaluator_factory.py
GuillaumeDSM: I put it there because service feeds are used by evaluators and should be created and started at a specific time to allow evaluators to register first. We can put it in a dedicated file and call it from ```evaluator_factory.py```

octobot/exchange_factory.py
Herklos: Ok be here you are making no difference between a backtesting bot and a classic one. And I don't think  we need a list of exchange_manager objects but only a list of exchange ids.

octobot/interface_factory.pxd
Herklos: Yes you are right in other interfaces there are some `_list`. We should change all of it in another PR.

octobot/evaluator_factory.py
Herklos: Yes I think we should put this in another file. But I don't know if it's better to call in from `evaluator_factory.py` or `octobot.py`.

start.py
```diff
@@ -168,10 +178,10 @@ def start_octobot(starting_args):
                     # In those cases load OctoBot
                     from octobot.octobot import OctoBot
 
-                    # TelegramApp.enable(config, not starting_args.no_telegram)
-                    # WebService.enable(config, not starting_args.no_web)
+                    _disable_interface_from_param("telegram", starting_args.no_telegram, logger)
+                    _disable_interface_from_param("web", starting_args.no_web, logger)
```
Herklos: :+1:

octobot/service_feed_factory.py
```diff
@@ -0,0 +1,42 @@
...
+        self.service_feeds = []
+
+    async def initialize(self):
+        service_feed_factory = create_service_feed_factory(self.octobot.config, self.octobot.async_loop)
+        self.service_feeds = [service_feed_factory.create_service_feed(feed)
+                              for feed in service_feed_factory.get_available_service_feeds()]
+
+    async def create(self):
+        for feed in self.service_feeds:
+            if not await start_service_feed(feed, False):
```
Herklos: :+1:

PR #1020, opened 2020-01-05
[EvaluatorFactory] Use new exchanges get all method

Herklos:
Requires https://github.com/Drakkar-Software/OctoBot-Trading/pull/60


Herklos:
> An API call would be even clearer but it's much better already, we can add this once the service branches are merged

Yes we should migrate this method in the future trading API.

PR #1023, opened 2020-01-11
Add metrics management

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Evaluators/pull/9
requires https://github.com/Drakkar-Software/OctoBot-Trading/pull/30
requires https://github.com/Drakkar-Software/OctoBot/pull/1009
requires https://github.com/Drakkar-Software/OctoBot-Trading/pull/68

### DIFF
--- a/tools/metrics/metrics_fields.py
+++ b/tools/metrics/metrics_fields.py
@@ -13,28 +13,22 @@
 #
 #  You should have received a copy of the GNU Lesser General Public
 #  License along with this library.
-from octobot_backtesting.api.backtesting import is_backtesting_enabled
-from tools.metrics.metrics_manager import MetricsManager
+from enum import Enum
 
 
-from tools import get_logger
-
-
-class Initializer:
-    """Initializer class:
-    - Initialize services, constants and tools
-    """
-
-    def __init__(self, octobot):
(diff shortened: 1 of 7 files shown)


tools/metrics/metrics_fields.py
```diff
@@ -0,0 +1,34 @@
...
+    UP_TIME = "uptime"
+    SIMULATOR = "simulator"
+    TRADER = "trader"
+    EVAL_CONFIG = "evalconfig"
+    PAIRS = "pairs"
+    EXCHANGES = "exchanges"
+    NOTIFICATIONS = "notifications"
+    TYPE = "type"
+    PLATFORM = "platform"
+    REFERENCE_MARKET = "referencemarket"
```
Herklos: Why not `reference_market` ?

GuillaumeDSM:
> About `class MetricsFields(Enum)`,
> Why not `reference_market`, `current_session`, `started_at` and `evalconfig` ?

When I developped the server side of metrics management, I didn't manage to handle _ properly, we can changed it but we need to update 0.3 as well, I don't think it's really important since it's just an internal data struct between the bot metrics service and the server side.

PR #1048, opened 2020-02-06
update trades_callback args

GuillaumeDSM:


tools/logger.py
```diff
@@ -81,8 +81,9 @@
                             f"{PrettyPrinter.portfolio_profitability_pretty_print(profitability, profitability_percent, 'USDT')}")
 
 
-async def trades_callback(exchange: str, exchange_id: str, symbol: str, trade):
-    BOT_CHANNEL_LOGGER.info(f"TRADES : EXCHANGE = {exchange} || SYMBOL = {symbol} || TRADE = {trade}")
+async def trades_callback(exchange: str, exchange_id: str, symbol: str, trade, old_trade: bool):
+    BOT_CHANNEL_LOGGER.info(f"TRADES : EXCHANGE = {exchange} || SYMBOL = {symbol} || TRADE = {trade} "
```
Herklos: What do you mean by "old_trade" ?

Herklos: It's when the trade is loaded when the bot is starting ?

GuillaumeDSM: It's a param that is in the trades channel callback

PR #1065, opened 2020-03-11
[ChannelsLogger] Fix position callback args 

Herklos:
Depends on https://github.com/Drakkar-Software/OctoBot-Trading/pull/127


Herklos:
Merged in https://github.com/Drakkar-Software/OctoBot/pull/1071

PR #1073, opened 2020-03-25
Package manager refactor 

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Commons/pull/28

When merged, to export local tentacles to an exisiting tentacles repo:
```bash
python start.py tentacles -e "/X/Y/Z/OctoBot-tentacles" -d ./tentacles
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -116,7 +116,9 @@ logs/OctoBot\.log\.*
 
 # Tentacles
 tentacles
-config\.json
+
+# User config
+user/
 
 *.csv
 *.ods
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -34,15 +34,6 @@ LOGS_FOLDER = "logs"
 CONFIG_KEY = "config"
 TENTACLES_SETUP_CONFIG_KEY = "tentacles_setup"
 
-
-# command line special args
-FORCE_ARG = "force"
-ALL_ARG = "all"
-HELP_ARG = "help"
-INSTALL_ARG = "install"
-UPDATE_ARG = "update"
-UNINSTALL_ARG = "uninstall"
-
 # terms of service
 CONFIG_ACCEPTED_TERMS = "accepted_terms"
 
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -25,9 +25,15 @@ LONG_VERSION = f"{SHORT_VERSION}{PATCH_VERSION}{VERSION_DEV_PHASE}{VERSION_PHASE
 # tentacles
 # TODO: use real link
 DEFAULT_TENTACLES_URL = f"https://raw.githubusercontent.com/GuillaumeDSM/test_tentacles2/master/tentacles.zip"
+DEFAULT_TENTACLES_PACKAGE_NAME = "OctoBot-Default-Tentacles"
 
 # logs
 LOGS_FOLDER = "logs"
+
+# config types keys
+CONFIG_KEY = "config"
+TENTACLES_SETUP_CONFIG_KEY = "tentacles_setup"
+
 
 # command line special args
 FORCE_ARG = "force"
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -22,8 +22,20 @@ VERSION_PHASE = "5"  # XX
 VERSION = f"{SHORT_VERSION}{VERSION_DEV_PHASE}{VERSION_PHASE}"
 LONG_VERSION = f"{SHORT_VERSION}{PATCH_VERSION}{VERSION_DEV_PHASE}{VERSION_PHASE}"
 
+# tentacles
+# TODO: use real link
+DEFAULT_TENTACLES_URL = f"https://raw.githubusercontent.com/GuillaumeDSM/test_tentacles2/master/tentacles.zip"
+
 # logs
 LOGS_FOLDER = "logs"
+
+# command line special args
+FORCE_ARG = "force"
+ALL_ARG = "all"
+HELP_ARG = "help"
+INSTALL_ARG = "install"
+UPDATE_ARG = "update"
+UNINSTALL_ARG = "uninstall"
 
 # terms of service
 CONFIG_ACCEPTED_TERMS = "accepted_terms"
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -22,9 +22,14 @@ VERSION_PHASE = "5"  # XX
 VERSION = f"{SHORT_VERSION}{VERSION_DEV_PHASE}{VERSION_PHASE}"
 LONG_VERSION = f"{SHORT_VERSION}{PATCH_VERSION}{VERSION_DEV_PHASE}{VERSION_PHASE}"
 
+# OctoBot urls
+OCTOBOT_ONLINE = "https://www.tentacles.octobot.online"
+REPOSITORY = "repository"
+TENTACLES_REPOSITORY = "tentacles"
+RELEASES = "releases"
+
 # tentacles
-# TODO: use real link
-DEFAULT_TENTACLES_URL = f"https://raw.githubusercontent.com/GuillaumeDSM/test_tentacles2/master/tentacles.zip"
+DEFAULT_TENTACLES_URL = f"{OCTOBOT_ONLINE}/{REPOSITORY}/{TENTACLES_REPOSITORY}/{RELEASES}/{SHORT_VERSION}.zip"
 DEFAULT_TENTACLES_PACKAGE_NAME = "OctoBot-Default-Tentacles"
 
 # logs
--- a/octobot/configuration_manager.pxd
+++ b/octobot/configuration_manager.pxd
@@ -16,7 +16,14 @@
 #  License along with this library.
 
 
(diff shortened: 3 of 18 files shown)


octobot/evaluator_factory.py
```diff
@@ -33,41 +32,20 @@ def __init__(self, octobot):
...
-        self.symbol_evaluator_list = {}
-        self.cryptocurrency_evaluator_list = {}
-        self.social_eval_tasks = []
-        self.real_time_eval_tasks = []
-
-        self.service_feed_list = []
-
-    async def initialize(self):
-        self.matrix_id = await initialize_evaluators(self.octobot.config)
+    async def initialize(self, tentacles_setup_config):
```
Herklos: :+1:

octobot/configuration_manager.py
```diff
@@ -0,0 +1,37 @@
...
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+#  Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser General Public
+#  License along with this library.
+import copy
+
+
+class ConfigurationManager:
```
Herklos: :+1:

octobot/configuration_manager.pxd
```diff
@@ -0,0 +1,29 @@
...
+
+cdef class ConfigurationManager:
+    cdef dict configuration_elements
+
+    cpdef void add_element(self, str key, object element)
+    cpdef object get_edited_config(self, str key)
+    cpdef object get_startup_config(self, str key)
+
+cdef class ConfigurationElement:
+    cdef public object config
```
Herklos: `config`, `startup_config`, `edited_config` are dict, aren't they ?

tools/commands.py
```diff
@@ -46,23 +52,66 @@ def data_collector(config, catch=True):
...
+        else:
+            await update_tentacles(command_args, DEFAULT_TENTACLES_URL, aiohttp_session=aiohttp_session)
+    elif uninstall:
+        if all:
+            await uninstall_all_tentacles(DEFAULT_TENTACLES_URL, use_confirm_prompt=force)
+        else:
+            await uninstall_tentacles(command_args, DEFAULT_TENTACLES_URL, use_confirm_prompt=force)
+
+
+def _print_tentacle_manager_help():
```
Herklos: I think this should be defined in a tentacle manager entrypoint

tools/commands.py
```diff
@@ -46,23 +52,66 @@ def data_collector(config, catch=True):
...
 
-def package_manager(config, commands, catch=False, force=False, default_git_branch=TENTACLES_DEFAULT_BRANCH):
+def _check_arg_and_cleanup(command_args, argument):
+    if argument in command_args:
+        command_args.pop(command_args.index(argument))
+        return True
+    return False
+
+
+async def _handle_package_manager_command(command_args, aiohttp_session=None):
```
Herklos: I think this should be defined in a tentacle manager entrypoint

octobot/configuration_manager.pxd
GuillaumeDSM: right

tools/commands.py
GuillaumeDSM: I was thinking about it and ended up adding it in OctoBot because it's using OctoBot's command line, like this tentacles-manager is just a lib and the entrypoint is only it's caller. You think we should share this command line management with the tentacles-manager ?

tools/commands.py
GuillaumeDSM: Same as the next answer, it depends if we want to share the command line responsibility with the tentacle manager or not. I think it's better to expose the API and leave the choice to the caller to use a CLI or not and how the CLI should behave

tools/commands.py
Herklos: it seemed to me that we had talked about the possibility of using the Tentacle Manager standalone. So with a CLI.

tools/logger.py
```diff
@@ -25,6 +32,33 @@
 BOT_CHANNEL_LOGGER = get_logger("OctoBot Channel")
 
 
+def _log_uncaught_exceptions(ex_cls, ex, tb):
+    logging.exception(''.join(traceback.format_tb(tb)))
+    logging.exception('{0}: {1}'.format(ex_cls, ex))
+
+
+def init_logger():
```
Herklos: Good idea to move this here :+1:

start.py
```diff
@@ -303,10 +254,21 @@ def main(args=None):
...
-    args = parser.parse_args(args)
+    # add sub commands
+    subparsers = parser.add_subparsers(title="Other commands")
 
-    start_octobot(args)
+    # tentacles manager
+    tentacles_parser = subparsers.add_parser("tentacles", help='Calls OctoBot tentacles manager.\n'
+                                                               'Use "tentacles --help" to get the '
+                                                               'tentacles manager help.')
+    register_tentacles_manager_arguments(tentacles_parser)
```
Herklos: :+1:

octobot/exchange_factory.py
```diff
@@ -49,12 +47,13 @@ def __init__(self, octobot, ignore_config=False):
...
         self.available_exchanges = ccxt.exchanges
 
-    async def create(self):
+    async def create(self, tentacles_setup_config):
         if self.octobot.config[CONFIG_EXCHANGES]:
             for exchange_class_string in self.octobot.config[CONFIG_EXCHANGES]:
                 if exchange_class_string in self.available_exchanges:
                     exchange_builder = create_exchange_builder(self.octobot.config, exchange_class_string) \
                                           .has_matrix(self.octobot.evaluator_factory.matrix_id) \
+                                          .use_tentacles_setup_config(tentacles_setup_config) \
```
Herklos: As the above lines we can have : `self.octobot.tentacles_setup_config` instead of `tentacles_setup_config` param

.gitignore
```diff
@@ -127,3 +127,4 @@ log
 logs/states/trading_state_history.json
 backtesting/data/
 cython_debug/
+user/tentacles_config/
```
Herklos: I think we should just ignore `user` (by removing the `config.json` ignore above and the `tentacles_config/`).

config/__init__.py
```diff
@@ -22,9 +22,18 @@
 VERSION = f"{SHORT_VERSION}{VERSION_DEV_PHASE}{VERSION_PHASE}"
 LONG_VERSION = f"{SHORT_VERSION}{PATCH_VERSION}{VERSION_DEV_PHASE}{VERSION_PHASE}"
 
+# tentacles
+# TODO: use real link
+DEFAULT_TENTACLES_URL = f"https://raw.githubusercontent.com/GuillaumeDSM/test_tentacles2/master/tentacles.zip"
```
Herklos: Should we use `www.tentacles.octobot.online` here ?

.gitignore
GuillaumeDSM: right, it's better

config/__init__.py
GuillaumeDSM: yes, when it will be available (hopefully tonight)

config/__init__.py
```diff
@@ -22,9 +22,14 @@
 VERSION = f"{SHORT_VERSION}{VERSION_DEV_PHASE}{VERSION_PHASE}"
 LONG_VERSION = f"{SHORT_VERSION}{PATCH_VERSION}{VERSION_DEV_PHASE}{VERSION_PHASE}"
 
+# OctoBot urls
+OCTOBOT_ONLINE = "https://www.tentacles.octobot.online"
```
Herklos: :+1:

octobot/octobot.py
```diff
@@ -96,16 +101,23 @@ def __init__(self, config, ignore_config=False, reset_trading_history=False):
     async def _post_initialize(self):
         self.initialized = True
 
-        # update startup_config and edited_config now that config contains all necessary info
```
Herklos, line 99: :+1:

octobot/octobot_api.py
```diff
@@ -31,14 +32,23 @@ def get_global_config(self) -> dict:
...
 
     def get_startup_config(self) -> dict:
-        return self._octobot.startup_config
+        return self._octobot.get_startup_config(CONFIG_KEY)
 
     def get_edited_config(self) -> dict:
-        return self._octobot.edited_config
+        return self._octobot.get_edited_config(CONFIG_KEY)
+
+    def get_startup_tentacles_config(self) -> dict:
```
Herklos, line 40: :+1:

requirements.txt
```diff
@@ -3,7 +3,7 @@ cython==0.29.14
 
 # Drakkar-Software requirements
 OctoBot-Commons>=1.2.1, <1.3
-OctoBot-Tentacles-Manager==1.0.12
+OctoBot-Tentacles-Manager>=2.0.0, <2.1
```
Herklos, line 6: :tada:

PR #1076, opened 2020-04-05
Tentacles migration 

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot/pull/1073


setup.cfg
```diff
@@ -17,4 +17,5 @@ pep8ignore =
     E241    # multiple spaces after ','
     E226    # missing white space around arithmetic operator
     E222    # multiple spaces after operator
+    W391    # blank line at the end of file
```
Herklos, line 20: Why ignoring ?

GuillaumeDSM: tentacles package manager is not creating eof blank lines in init files, it makes pytest fail on this lint check

Herklos: ok :smile:

PR #1077, opened 2020-04-06
[Dockerfile] Update

Herklos:


Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN pip install --upgrade pip \
     && pip install --prefer-binary -r requirements.txt \
     && python setup.py install
 
-FROM python:3.7-slim
+FROM python:3.8-alpine
```
GuillaumeDSM, line 15: FROM python:3.8-slim AS base
and
FROM python:3.8-alpine
?

Herklos: No I'll find you something to explain this

Herklos: https://docs.docker.com/develop/develop-images/multistage-build/#name-your-build-stages

PR #1080, opened 2020-04-06
Tentacles migration adaptations 

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Tentacles-Manager/pull/18
requires https://github.com/Drakkar-Software/OctoBot-Services/pull/14
requires https://github.com/Drakkar-Software/OctoBot-Evaluators/pull/42

### DIFF
--- a/octobot/octobot.pxd
+++ b/octobot/octobot.pxd
@@ -34,6 +34,7 @@ cdef class OctoBot:
 
     cdef public dict tools
     cdef public dict config
+    cdef public str bot_id
 
     cdef object _aiohttp_session
     cdef public object community_handler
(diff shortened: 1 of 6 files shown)


octobot/octobot.pxd
```diff
@@ -34,6 +34,7 @@ cdef class OctoBot:
 
     cdef public dict tools
     cdef public dict config
+    cdef public str id
```
Herklos, line 37: I don't think `id` is a good idea, it's a [reserved python keyword](https://docs.python.org/3.5/library/functions.html#id)

GuillaumeDSM: right

PR #1106, opened 2020-04-15
update strategy test framework 

GuillaumeDSM:


tests/functional_tests/strategy_evaluators_tests/abstract_strategy_test.py
```diff
@@ -42,62 +44,47 @@ class AbstractStrategyTest(AbstractBacktestingTest, ABC):
...
 
-        assert run_results[0] >= profitability
+    def _handle_results(self, independent_backtesting, profitability):
+        exchange_manager_ids = get_independent_backtesting_exchange_manager_ids(independent_backtesting)
+        for exchange_manager in get_exchange_managers_from_exchange_ids(exchange_manager_ids):
+            _, run_profitability, _, market_average_profitability, _ = get_profitability_stats(exchange_manager)
+            # convenient for building tests
+            print(f"results: run profitability {run_profitability} market profitability: {market_average_profitability}"
+                  f" expected: {profitability} => {run_profitability >= profitability}")
+            assert run_profitability >= profitability
```
Herklos, line 61: We accept superior profitability ?

GuillaumeDSM: that's how it was previously, actually it might be bettter with a 
```python
profitability * 1.1 =>  run_profitability >= profitability
```
What do you think ?

Herklos: In fact I don't understand why it's not `assert run_profitability == profitability` ?

GuillaumeDSM: Not sure why it was like this in 0.3, let's put a == for now and be optimistic :)

PR #1111, opened 2020-04-23
Strategy update 

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Evaluators/pull/48


requirements.txt
```diff
@@ -1,5 +1,5 @@
 # Setup requirements
-cython==0.29.16
+cython==0.29.17
```
Herklos, line 2: :+1:  Don't forget to also update `setup.py`

Herklos: Don't forget to also update Commons to 1.3.12 :wink:

Herklos: And channels to 1.3.25

PR #1147, opened 2020-05-02
[Version] v0.4.0-alpha10 

Herklos:
Depends on https://github.com/Drakkar-Software/OctoBot-Trading/pull/173
Depends on https://github.com/Drakkar-Software/OctoBot-Backtesting/pull/59
Depends on https://github.com/Drakkar-Software/OctoBot-Evaluators/pull/62

### Added
- [Backtesting] Synchronized channel support

### Updated
- [Tests] strategy and TA framework
- [Tests] trading mode framework migration
- [Interfaces & Notifications] Migration follow up
- [Backtesting] Only one instance is created

### Fixed
- [Tests] Stress test timeout 
- [Startup] Fix api calls

PR #1151, opened 2020-05-04
fix imports and tentacles manager return code 

GuillaumeDSM:


GuillaumeDSM:
to be included in https://github.com/Drakkar-Software/OctoBot/pull/1147

PR #1157, opened 2020-05-07
fix memory check and add tentacle memory check util function

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Commons/pull/66
requires https://github.com/Drakkar-Software/OctoBot/pull/1154

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,12 +2,12 @@
 cython==0.29.17
 
 # Drakkar-Software requirements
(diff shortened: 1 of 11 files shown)


requirements.txt
```diff
@@ -5,7 +5,7 @@ cython==0.29.17
 OctoBot-Commons>=1.3.18, <1.4
```
Herklos, line 5: Should also use new commons version I think

PR #1162, opened 2020-05-09
Fix metrics name

Herklos:


GuillaumeDSM:
Looks like ```cdef void _init_metrics(self)``` remains at least on one pyd file

PR #1177, opened 2020-05-12
fix trading mode init

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Trading/pull/196


Herklos:
I don't understand the goal of this PR

GuillaumeDSM:
> I don't understand the goal of this PR

Trading modes advanced classes are initialized in exchange_builder now, that's an issue because if 2 or more exchanges are created, advanced classes are duplicated also. Trading mode advanced classes should be managed independently from exchange creation

Herklos:
> > I don't understand the goal of this PR
> 
> Trading modes advanced classes are initialized in exchange_builder now, that's an issue because if 2 or more exchanges are created, advanced classes are duplicated also. Trading mode advanced classes should be managed independently from exchange creation

Ok I understand. Can we move that to the Tentacle Manager ?

PR #1219, opened 2020-05-28
Remove sudo warning

Herklos:


octobot/__init__.py
```diff
@@ -37,16 +37,6 @@
...
 
-# if compatible version, can proceed with imports
-from octobot_commons.enums import PlatformsName
-from octobot_commons.os_util import get_os
-from octobot_commons.logging.logging_util import get_logger
-
-# check sudo rights
-if get_os() is not PlatformsName.WINDOWS and os.getuid() == 0:
-    get_logger("PythonChecker").warning("OctoBot is started with admin / sudo rights that are not required, "
-                                        "please check you starting command ")
```
GuillaumeDSM, line 48: Why ?

Herklos: It's always print with docker... and it's useless

PR #1247, opened 2020-06-16
updte azure readme badge

GuillaumeDSM:
It looks like the link is still not working even though it's the azure auto link, apparently it's the same for other octobot projects :o

This PR might be useless then


GuillaumeDSM:
useless fix, azure link is broken for now

PR #1264, opened 2020-07-23
fix octobot stop method

GuillaumeDSM:


azure-pipelines.yml
```diff
@@ -54,9 +54,9 @@ jobs:
       {
         $Env:TENTACLES_URL_SUBCATEGORY = ""
         If ($env:build_sourceBranch -Contains "refs/tags/") {
-          $Env:TENTACLES_URL_TAG = "latest"
-        } else {
           $Env:TENTACLES_URL_TAG = ""
+        } else {
+          $Env:TENTACLES_URL_TAG = "latest"
```
Herklos, line 59: :+1:

tests/functional_tests/launch_test/launch_test.py
```diff
@@ -33,4 +35,27 @@ async def test_run_bot():
     bot.task_manager.init_async_loop()
     await start_bot(bot, init_logger())
     await asyncio.sleep(10)
-    bot.task_manager.stop_tasks()
```
Herklos, line 36: Why don't we use `stop_tasks()` anymore ?

GuillaumeDSM: to be able to await the stop (stop tasks is not a coroutine and is using call_soon which makes it not awaitable)

Herklos: Why not just calling `task_manager.stop_tasks` in `async def stop_bot(bot)`  ?

PR #1362, opened 2020-10-29
raise asyncio.Timeout errors

GuillaumeDSM:


tests/functional_tests/backtesting_tests/backtesting_test.py
```diff
@@ -26,6 +26,8 @@
 from octobot.backtesting.abstract_backtesting_test import DATA_FILES
 from octobot_trading.api.trades import get_trade_history
 from tests.test_utils.bot_management import run_independent_backtesting
+# force tentacles import since it's required for backtesting
+import tentacles
```
Herklos, line 30: I don't understand why we need to do this now when we didn't before.

GuillaumeDSM: It's an issue we already had before, it's impossible to run backtesting tests only if we don't put this import since no tentacle import is done anywhere else (was previously done in octobot-evaluators which was an issue)

GuillaumeDSM:
fix in https://github.com/Drakkar-Software/OctoBot/pull/1366

PR #1368, opened 2020-10-31
add community authentication handler

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot/pull/1387


requirements.txt
```diff
@@ -15,3 +15,4 @@ colorlog==4.2.1
 yarl==1.1.0
 idna<2.9,>=2.5
 aiohttp==3.7.2
+requests==2.24.0
```
Herklos, line 18: Why not using aiohttp ?

octobot/community/authentication.py
```diff
@@ -0,0 +1,177 @@
...
+        self.authentication_url = authentication_url
+        self.refresh_token = None
+        self.edited_config = config
+
+        self._auth_token = None
+        self._expire_at = None
+        self._session = requests.Session()
+        self._cache = {}
+
+        if None not in (username, password):
```
Herklos, line 52: I'm not a big fan of `None not in (username, password)`. I would prefer `username and password`

octobot/community/authentication.py
```diff
@@ -0,0 +1,177 @@
...
+            return self._cache[url]
+        else:
+            return self._session.get(url, params=params, **kwargs)
+
+    @authenticated
+    def post(self, url, data=None, json=None, **kwargs):
+        return self._session.post(url, data=data, json=json, **kwargs)
+
+    def is_logged_in(self):
+        return None not in (self._auth_token, self.refresh_token, self._expire_at)
```
Herklos, line 90: Same here

octobot/community/authentication.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def ensure_token_validity(self):
```
Herklos, line 92: :+1:

octobot/community/authentication.py
```diff
@@ -0,0 +1,177 @@
...
+            # try to login with saved token if available
+            self._try_auto_login()
+            if not self.is_logged_in():
+                # still not logged in: raise
+                raise AuthenticationRequired()
+        if time.time() >= self._expire_at:
+            self._refresh_auth()
+
+    def remove_login_detail(self):
+        self._update_login_config("")
```
Herklos, line 103: Why not passing no argument with `_update_login_config(self, value="")` ?

octobot/community/authentication.py
```diff
@@ -0,0 +1,177 @@
...
+            self._refresh_auth()
+        except FailedAuthentication:
+            self.logout()
+        except Exception as e:
+            self.logger.exception(e, True, f"Error when trying to refresh community login: {e}")
+
+    def _refresh_auth(self):
+        params = {
+            "refresh_token": self.refresh_token,
+            "grant_type": "refresh_token",
```
Herklos, line 130: Can we add a class constant for this two strings ?

tests/unit_tests/community/test_authentication.py
```diff
@@ -0,0 +1,343 @@
...
+import pytest
+import mock
+import requests
+import time
+
+import octobot.community as community
+import octobot_commons.config_manager
+import octobot_commons.constants as commons_constants
+
+AUTH_URL = "https://oh.fake/auth"
```
Herklos, line 25: Could we use `os.getenv("COMMUNITY_SERVER_URL", "https://oh.fake/auth")` to test with a real server ? (same for www.todo.com below)

tests/unit_tests/community/test_authentication.py
```diff
@@ -0,0 +1,343 @@
...
+            mock.patch.object(community.CommunityAuthentication, "_handle_auth_result", mock.Mock()) \
+            as auth_res_mock, \
+            mock.patch.object(requests, "post", mock.Mock(return_value="1")) as post_mock:
+        auth.login("username", "password")
+        reset_mock.assert_called_once()
+        post_mock.assert_called_once()
+        auth_res_mock.assert_called_once_with("1")
+
+
+def test_logout(auth):
```
Herklos, line 81: :+1:

tests/unit_tests/community/test_authentication.py
```diff
@@ -0,0 +1,343 @@
...
+    assert auth.is_logged_in() is False
+    auth._auth_token = "1"
+    auth.refresh_token = "1"
+    auth._expire_at = None
+    assert auth.is_logged_in() is False
+    auth._expire_at = "1"
+    assert auth.is_logged_in() is True
+
+
+def test_ensure_token_validity():
```
Herklos, line 216: :100:

Herklos (commented):
Very nice job :+1: Great test coverage :ok_hand:

octobot/community/authentication.py
GuillaumeDSM: The problem with this check is that it prevents the opportunity for an empty password / username auth

Herklos: Do you think we should able to use empty login username and password ?

octobot/community/authentication.py
GuillaumeDSM: Here `_update_login_config` is supposed to receive a value, "" is more of an exception than the default behavior

Herklos: Ok, in my opinion it's not very clear to say "update something with an empty value" to reset / remove a thing.

octobot/community/authentication.py
GuillaumeDSM: I just don't see why we should forbid

octobot/community/authentication.py
GuillaumeDSM: Right, I'll see how to make it clearer

Herklos (approved):

Herklos (approved):

GuillaumeDSM:
contained in https://github.com/Drakkar-Software/OctoBot/pull/1395

PR #1451, opened 2020-12-13
[Docker] Update Dockerfile and docker-compose.yml

Herklos:
- Dockerfile with new default volumes
- docker-compose.yml with default http port and new default volumes


GuillaumeDSM (approved):
👍

PR #1459, opened 2020-12-26
[CI] Enable fork to create PR and have CI running

Herklos:
Uses pull_request according to https://securitylab.github.com/research/github-actions-preventing-pwn-requests to run jobs securely

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,5 +1,10 @@
 name: OctoBot-CI
-on: push
+on:
+  push:
+    branches:
+      - 'master'
+      - 'dev'
+  pull_request:
 
 jobs:
   lint:


.github/workflows/main.yml
```diff
@@ -1,5 +1,10 @@
 name: OctoBot-CI
-on: push
+on:
+  push:
+    branches:
+      - 'master'
+      - 'dev'
+  pull_request:
```
GuillaumeDSM, line 7: is it fine if we keep it all in the same workflow ? I mean in the same .yml

Herklos: I don't understand your question, why spliting ?

GuillaumeDSM: In the github doc example, they split workflows into 2 files (on with access to secrets and the other without), I'm wondering if this is required or not in our case (to protect secrets)

Herklos: They said :
> To do this in a secure manner, the untrusted code must be handled via the pull_request trigger so that it is isolated in an unprivileged environment

And the first example explains that `pull_request` give no access to secrets : 
```
# read-only repo token
# no access to secrets
on:
  pull_request:
```
The second and third example are completly different from the first one.

I agree one point : we'll need additional changes to enable our CI to succeed without secrets access (we currently have some unecessary secrets that should be replaced by hardcoded values).

GuillaumeDSM: Right, what I'm wondering is if having both `push` and `pull_request` triggers on the same workflow is not making it so that `pull_request` get the `push` rights as well

GuillaumeDSM (approved):

PR #1460, opened 2020-12-26
[CI] Add docker run and check health

Herklos:
Depends on #1461


.github/workflows/main.yml
```diff
@@ -274,7 +283,7 @@ jobs:
             ${{ env.IMAGE }}:${{ env.VERSION }}
```
GuillaumeDSM, line 283: we don't push this anymore ?

Herklos: It's exactly the same behavior as before.

GuillaumeDSM (approved):

.github/workflows/main.yml
GuillaumeDSM: right, my bad !

PR #1462, opened 2020-12-26
[Docker] Fix tentacle tag arg

Herklos:


Dockerfile
```diff
@@ -1,5 +1,9 @@
+ARG TENTACLES_URL_TAG
+
 FROM python:3.8-slim-buster AS base
 
+ARG TENTACLES_URL_TAG
```
GuillaumeDSM, line 5: `ARG TENTACLES_URL_TAG`

twice in dockerfile ?

Herklos: It's mandatory for multistage builds

GuillaumeDSM (approved):

PR #1475, opened 2021-01-02
[CI] Add codeql workflow

Herklos:
Recommended by github in https://github.com/Drakkar-Software/OctoBot/security/code-scanning

### DIFF
--- a/.github/workflows/analyze.yml
+++ b/.github/workflows/analyze.yml
@@ -1,0 +1,56 @@
+name: "OctoBot-CodeQL"
+
+on:
+  push:
+    branches: [ master, dev ]
+  pull_request:
+    branches: [ master, dev ]
+  schedule:
+    - cron: '43 20 * * 4' # recommended schedule
+
+jobs:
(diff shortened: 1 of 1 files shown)


GuillaumeDSM (commented):
image
should we handle these warnings ?

Herklos:
> image
> should we handle these warnings ?

I don't know, you think we should install dependencies ?

Herklos:
> image
> should we handle these warnings ?

You right, there is an issue with dependencies installation. It's probably due to python version

GuillaumeDSM:
> > image
> > should we handle these warnings ?
> 
> You right, there is an issue with dependencies installation. It's probably due to python version

It looks like there is something with python2

Herklos:
Worked in https://github.com/Drakkar-Software/OctoBot/runs/1636945938. I'll now change event trigger to `branches: [ master, dev ]` only as recommended.

GuillaumeDSM (approved):
great !

PR #1484, opened 2021-01-05
[Docs] Add analytics

Herklos:


docs/source/conf.py
```diff
@@ -54,6 +54,7 @@
     "collapse_navigation": False,
     "display_version": True,
     "logo_only": False,
+    "analytics_id": "G-Y4FB265L3N",
```
GuillaumeDSM, line 57: where is it used ?

Herklos: Should be used by the theme here https://github.com/asteroid-team/asteroid_sphinx_theme/blob/master/asteroid_sphinx_theme/theme.conf#L8 but seems not to be working.

Herklos: I'll remove it.

GuillaumeDSM (approved):

PR #1595, opened 2021-03-14
[BacktestingTests] do not cythonize _handle_results to prevent warn

GuillaumeDSM:
fixes 
image

ex in: https://github.com/Drakkar-Software/OctoBot-Tentacles/runs/2106148765?check_suite_focus=true#step:7:221

### DIFF
--- a/octobot/backtesting/abstract_backtesting_test.pxd
+++ b/octobot/backtesting/abstract_backtesting_test.pxd
@@ -26,5 +26,4 @@ cdef class AbstractBacktestingTest:
                                         object tentacles_setup_config,
                                         dict config)
 
-    cdef void _handle_results(self, object independent_backtesting, object profitability)
     cdef void _register_only_strategy(self, object strategy_evaluator_class)


octobot/backtesting/abstract_backtesting_test.pxd
```diff
@@ -26,5 +26,4 @@ cdef class AbstractBacktestingTest:
                                         object tentacles_setup_config,
                                         dict config)
 
-    cdef void _handle_results(self, object independent_backtesting, object profitability)
```
Herklos, line 29: The "real" fix would have been to return something instead of void. But it also works like this.

GuillaumeDSM: Not sure how it would fix it with void, the issue is that local implementations (strategy tests) are not cythonized (which is normal) therefore the cythonized version of the method is used and it's what is raising this error (this error should not be raised)

Herklos: Cython is silencing exception when a cythonized method doesn't return anything. It's not related to how the method is called.

GuillaumeDSM: Yes but the issue is that the wrong method is called, it should not be called at all and the cython declaration is preventing this call

Herklos: Oh ok, understood

Herklos (approved):

PR #1612, opened 2021-03-23
Disable rust compile on dockerfile

GuillaumeDSM:
Based on https://github.com/Drakkar-Software/OctoBot/pull/1611
thanks to @gabriel-milan

PR #1616, opened 2021-03-25
[Docker] build arm images only on dev or tag pushes

GuillaumeDSM:
Goal:
Since armv7 build is more than 3h long, only build it on dev and tag pushes

Related to:
https://github.com/Drakkar-Software/OctoBot/pull/1612#issuecomment-807691963


Herklos (approved):

Herklos:
Indeed 😉

PR #1653, opened 2021-04-12
[Commands] Add tentacle self update

Herklos:


requirements.txt
```diff
@@ -16,4 +16,3 @@ yarl==1.1.0
 idna<2.9,>=2.5
 requests==2.25.1
 chardet<4.0
-packaging==20.9
```
GuillaumeDSM, line 19: we don't make any version comparison anymore in octobot ?

octobot/initializer.py
```diff
@@ -27,9 +29,12 @@ def __init__(self, octobot):
...
     async def create(self):
         # initialize tentacle configuration
-        tentacles_config_path = self.octobot.get_startup_config(constants.CONFIG_KEY, dict_only=False).\
-            get_tentacles_config_path()
+        config = self.octobot.get_startup_config(constants.CONFIG_KEY, dict_only=False)
+        tentacles_config_path = config.get_tentacles_config_path()
         self.octobot.tentacles_setup_config = tentacles_manager_api.get_tentacles_setup_config(tentacles_config_path)
 
+        # check if tentacles setup requires an update
+        await commands.update_tentacles_if_necessary(self.octobot.tentacles_setup_config, config)
```
GuillaumeDSM, line 37: Why here and not in cli ?

Herklos: There is no `tentacles_setup_config` in cli

requirements.txt
Herklos: It's a dependency of Tentacle-Manager now. It'll be installed with OctoBot-Tentacles-Manager==2.4.11

GuillaumeDSM: I'm wondering if we should freeze the version of `packaging` in octobot and have a range of versions in OctoBot-Tentacles-Manager, what do you think about it ? Considering it's required in both

Herklos: I don't understand why we should do that, but we can.

GuillaumeDSM: Not that it's necessary but in my opinion it would be cleaner to have it in both repo since both are importing this package directly (and removing it OctoBot-Tentacles-Manager for whatever reason if we don't need it anymore won't break OctoBot without us noticing it in OctoBot-Tentacles-Manager CI)

Herklos: We should then define a requirement policy, we don't do this for aiohttp and others. I agree that it would be cleaner, but it will be harder to maintain.

GuillaumeDSM (approved):

PR #1680, opened 2021-05-04
[Config] Use default profile when profile is not found

Herklos:
Depends on https://github.com/Drakkar-Software/OctoBot/pull/1674


octobot/configuration_manager.py
```diff
@@ -128,16 +129,18 @@ def init_config(
         raise Exception(f"Can't init config file {global_exception}")
 
 
-def init_default_profile():
-    profile_folder = os.path.join(common_constants.USER_PROFILES_FOLDER,
-                                  common_constants.DEFAULT_PROFILE)
-    if not os.path.exists(profile_folder):
-        os.makedirs(profile_folder)
```
GuillaumeDSM, line 135: when is the profile created now ?

Herklos: When installing base tentacle package

GuillaumeDSM (approved):

PR #1713, opened 2021-05-23
[Loggers] Disable DEBUG logs by default. Reactivable with ENV var

Herklos:
Required to prevent spam with websockets


GuillaumeDSM (approved):
good idea

PR #1726, opened 2021-06-02
[CLI] Fix crash when using default config

Herklos:


octobot/cli.py
```diff
@@ -54,6 +54,11 @@ def update_config_with_args(starting_args, config: configuration.Configuration,
             "Can't start backtesting without the octobot_backtesting package properly installed.")
         raise e
 
+    if common_constants.CONFIG_TRADER not in config.config:
+        config.config[common_constants.CONFIG_TRADER] = {}
+    if common_constants.CONFIG_SIMULATOR not in config.config:
+        config.config[common_constants.CONFIG_SIMULATOR] = {}
```
GuillaumeDSM, line 60: On 1st start, this is not set because it's fetched from profiles which are created a bit later (in `_load_or_create_tentacles`), Is the real issue that profiles are sometimes not loaded ?

Herklos: That's right, I'll investigate more to know what's happening with profiles

PR #1731, opened 2021-06-05
[Commands] Catch ValueError when creating signal handler not in main Thread

Herklos:


octobot/commands.py
```diff
@@ -132,7 +132,10 @@ def _signal_handler(_, __):
 
 def run_bot(bot, logger):
     # handle CTRL+C signal
-    signal.signal(signal.SIGINT, _signal_handler)
+    try:
+        signal.signal(signal.SIGINT, _signal_handler)
+    except ValueError as e:
+        logger.error(f"Can't setup signal handler : {e}")
```
GuillaumeDSM, line 138: Should this be a warning ?

Herklos: Right

GuillaumeDSM (approved):

PR #1798, opened 2021-08-12
[Logger] Disable chan logger when ENABLE_DEBUG_LOGS is False

Herklos:


octobot/logger.py
```diff
@@ -157,6 +158,8 @@ async def init_exchange_chan_logger(exchange_id):
 
 
 async def init_evaluator_chan_logger(matrix_id: str):
+    if os.getenv(constants.ENV_ENABLE_DEBUG_LOGS, False):
+        return
```
GuillaumeDSM, line 162: Maybe we should let orders / position / liquidation / trades channels

Herklos: I don't know. Orders are still logging 
image

I don't think the channel subscription is the right thing to keep.

GuillaumeDSM (approved):

GuillaumeDSM (approved):

PR #1827, opened 2021-09-12
Update Readme

valouvaliavlo:
Thank you for sending your pull request. 
But first, have you included unit tests, and is your code PEP8 conformant? [More details](https://github.com/Drakkar-Software/OctoBot/blob/dev/CONTRIBUTING.md)

## Summary
Explain in one sentence the goal of this PR

Solve the issue: #___

## Changelog

  - _____
  - _____
  - _____

## What's new?
*Explain in details what this PR solve or improve.*

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Octobot's main feature is **evolution** : you can [install](https://docs.octobot
 ## Installation
 OctoBot's installation is **very simple**... because **very documented** ! See the [installation guides](https://www.octobot.online/guides/#installation) for more info.
 
-#### With executable
+#### [With executable](https://docs.octobot.online/installation/with-binary)
 Follow the [2 steps installation guide](https://www.octobot.online/executable_installation/) 
 
 In short:
--- a/README.md
+++ b/README.md
@@ -60,6 +60,14 @@ docker run -itd --name OctoBot -p 80:5001 -v $(pwd)/user:/octobot/user -v $(pwd)
 ```
 And then open [http://localhost](http://localhost).
 
+#### [With pip](https://docs.octobot.online/installation/with-pip)
+
+In short :
+```
+pip install OctoBot>=0.4.0b14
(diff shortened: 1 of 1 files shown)


GuillaumeDSM (approved):

Herklos (approved):

README.md
```diff
@@ -60,7 +60,15 @@ docker run -itd --name OctoBot -p 80:5001 -v $(pwd)/user:/octobot/user -v $(pwd)
 ```
 And then open [http://localhost](http://localhost).
 
-#### [With python sources](https://docs.octobot.online/installation/with-docker)
+#### [With pip](https://docs.octobot.online/installation/with-pip)
+
+In short :
+```
+pip install OctoBot>=0.4.0b14
```
GuillaumeDSM, line 67: actually, if we use `pip install -U OctoBot` we don't need the version

Herklos: Does it work when it is not already installed ?

valouvaliavlo: it's 0.3.9 which is installed without version of specified

valouvaliavlo: https://pypi.org/project/OctoBot/

Herklos: Right, that's why I said we''ll have to update this line at each update. I don' t know an option to install the latest pre-release version

GuillaumeDSM: Then I think `OctoBot>=0.4.0b14` is enough not to install 0.3.9, we don't need to update this number alongside the version number in new versions

Herklos:
```
~ $ pip3 install -U OctoBot
Collecting OctoBot
  Using cached OctoBot-0.3.9-py2.py3-none-any.whl (682 kB)
```

I agree OctoBot>=0.4.0b14 can do the job. I think we have to stop using pre release version form.

GuillaumeDSM: right

PR #1852, opened 2021-10-03
WIP [Python3.9] prevent interpreter shutdown error

GuillaumeDSM:
Fixes the interpreter shutdown error however prevents the bot from stopping, hence the WIP

It looks like the issue is that threadpoolexecutors are not daemons anymore and need to be shutdown (https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.shutdown)


Herklos:
Now in the code base.

PR #1992, opened 2022-07-14
Allow custom exchange type and adding limited random optimizer runs 

techfreaque:
Allow custom exchange type and adding limited random optimizer runs


octobot/backtesting/independent_backtesting.py
```diff
@@ -65,6 +65,10 @@ def __init__(self, config,
         self.optimizer_id = None
         self.backtesting_id = None
         self._init_default_config_values()
+        try:
+            self.backtesting_config["exchange_type"] = config["exchange_type"]
```
GuillaumeDSM, line 69: Could we use a constant here (like `CONFIG_EXCHANGE_TYPE`) please ?

octobot/backtesting/octobot_backtesting.py
```diff
@@ -217,6 +217,17 @@ async def _create_service_feeds(self):
                                   f"might not work properly")
 
     async def _init_exchanges(self):
+        if self.backtesting_config['exchange_type'] != "current_profile":
+            for exchange in self.exchange_type_by_exchange:
+                if self.backtesting_config['exchange_type'] == "SPOT":
```
GuillaumeDSM, line 222: We should use octobot_trading enums `ExchangeTypes` values

octobot/strategy_optimizer/strategy_design_optimizer.py
```diff
@@ -702,6 +705,10 @@ def _generate_runs(self):
             if self._is_run_allowed(run)
         }
         if runs:
+            temp_runs = list(runs.values())
+            random.shuffle(temp_runs)
+            temp_runs = temp_runs[:self.queue_size]
+            runs = dict(zip(runs, temp_runs))
```
GuillaumeDSM, line 711: 👍

octobot/config/profile_schema.json
```diff
@@ -138,6 +138,10 @@
               "type": "number",
               "minimum": 0,
               "maximum": 1
+            },
+            "current-bot-recording-id": {
+              "type": "integer",
+              "minimum": 1
```
GuillaumeDSM, line 144: what is a bot recording id and what is it used for ?

techfreaque: its used to switch the live run database (for plots and so on).
Its similar to the optimizer campaign id, but only for live trading

octobot/backtesting/independent_backtesting.py
techfreaque: thats already done

octobot/config/profile_schema.json
GuillaumeDSM: Therefore I guess it's local to a profile and used for display only, right ? In this case I think it would be better to put it in the strategy designer configuration since it's only used by this (unless I'm missing something there).

octobot/backtesting/octobot_backtesting.py
techfreaque: the current code uses the enums, but still the ifs. How would you remove the ifs?

octobot/config/profile_schema.json
techfreaque: its also used for the portfolio value history json file that gets written every day, even without the strategy designer I think

PR #2026, opened 2022-09-07
[Beta] handle beta tentacles 

GuillaumeDSM:


octobot/configuration_manager.py
```diff
@@ -143,14 +144,20 @@ def set_default_profile(config, from_default_config_file=constants.DEFAULT_CONFI
     config.save()
 
 
-def get_default_tentacles_url(version=None):
-    if version is None:
+def get_default_tentacles_url(version=None, use_beta_tentacles=False):
```
Herklos, line 147: Couldn't we use `version="beta"` instead of `use_beta_tentacles`?

GuillaumeDSM: Right, it will be simpler, I added it because it uses another repository but in the end the user of this method doesn't have to know it, better keeping it simple

octobot/configuration_manager.py
```diff
@@ -144,13 +145,19 @@ def set_default_profile(config, from_default_config_file=constants.DEFAULT_CONFI
 
 
 def get_default_tentacles_url(version=None):
+    beta_tentacle_bundle = os.getenv(constants.ENV_BETA_TENTACLES_PACKAGE_NAME, constants.BETA_TENTACLE_PACKAGE_NAME)
+    # use beta tentacles repository for beta tentacles
+    tentacles_repository = \
+        os.getenv(constants.ENV_BETA_TENTACLES_REPOSITORY, constants.BETA_TENTACLES_REPOSITORY) \
+        if version == beta_tentacle_bundle else \
```
Herklos, line 152: 👍

Herklos (approved):
👍

PR #2041, opened 2022-09-17
[Deploy] Add Kubernetes deployment file

Herklos:


octobot.yaml
```diff
@@ -0,0 +1,122 @@
...
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 100Mi
+status: {}
+
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
```
Herklos, line 96: We may switch to https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/ for `tentacles` volume

octobot.yaml
```diff
@@ -0,0 +1,105 @@
...
+            - mountPath: /octobot/backtesting
+              name: octobot-backtesting
+            - mountPath: /octobot/tentacles
+              name: octobot-tentacles
+            - mountPath: /octobot/user
+              name: octobot-user
+      restartPolicy: Always
+      volumes:
+        - name: octobot-logs
+          emptyDir: {}
```
GuillaumeDSM, line 54: does that mean we not save logs for an execution to another ?

Herklos: Logs will also be retrieved with the console output of the containers so I don't think we need to store them.

GuillaumeDSM: The console output has to be as detailed as the logfiles one for kub in this case 👍

GuillaumeDSM (approved):

octobot.yaml
Herklos: I'm not sure we want to store hundreds of megabytes per bot per day with the DEBUG level.

GuillaumeDSM: It should not be that much, it's reduced by a lot now. For example right now I have a 1h runing bot with 31Ko logs. It's to be analysed if we need debug logs or not I think in cloud bots

Herklos: I think it would be easier to have the logs collected in the output of the console even if it means modifying the logs level when we are in the cloud. It will be much simpler rather than going to recover log files if needed

kaovilai:
Had just tried to run this.. but I think the IS_CLOUD_ENV is [triggering unimplemented behavior](https://github.com/Drakkar-Software/OctoBot-Commons/pull/260/files#diff-cee2b7263dd9741d9ee9a2d6bf9761d74cc862fc992144a8fffe30519dbed51cR77).

Herklos:
> Had just tried to run this.. but I think the IS_CLOUD_ENV is [triggering unimplemented behavior](https://github.com/Drakkar-Software/OctoBot-Commons/pull/260/files#diff-cee2b7263dd9741d9ee9a2d6bf9761d74cc862fc992144a8fffe30519dbed51cR77).

Hi! Yes we are working on the cloud env, it may not be stable for now.

ruidazeng (approved):

rstuhlmuller:
I have created my own that I am using here: https://github.com/rstuhlmuller/octobot-deploy

kaovilai (approved):

PR #2105, opened 2022-11-12
Limits

GuillaumeDSM:


octobot/limits.py
```diff
@@ -0,0 +1,73 @@
...
+                crypto_currency_data[common_constants.CONFIG_CRYPTO_PAIRS] = updated_symbols
+        if len(trading_api.get_config_symbols(dict_config, True)) > limit:
+            logger.error(f"Too many trading pairs, maximum allowed is {limit}. Enabled : {enabled_symbols}")
+
+
+def apply_config_limits(configuration):
+    logger = logging.get_logger("ConfigurationLimits")
+    try:
+        _apply_exchanges_limits(configuration.config, logger, constants.MAX_ALLOWED_EXCHANGES)
+        _apply_symbols_limits(configuration.config, logger, constants.MAX_ALLOWED_SYMBOLS)
```
Herklos, line 71: Shouldn't we check here if we have to call the apply methods. For example by adding:
```
if constants.MAX_ALLOWED_EXCHANGES != constants.UNLIMITED_ALLOWED
```
before the first method

Herklos (approved):
👌

PR #2115, opened 2022-11-20
Fix cannot import name 'Log' from 'distutils.log'

Herklos:
Inspired from https://github.com/pypa/setuptools/issues/3693 and https://github.com/numpy/numpy/issues/22623

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,6 +18,7 @@ idna<2.9,>=2.5
 requests==2.25.1
 packaging==21.3
 python-dotenv==0.21.0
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -18,6 +18,7 @@ idna<2.9,>=2.5
 requests==2.25.1
 packaging==21.3
 python-dotenv==0.21.0
+setuptools<65.6
```
GuillaumeDSM, line 21: can we add a comment to remember why we did this and remove it when fixed plz ?

Herklos: Right 👍

GuillaumeDSM (approved):

PR #2126, opened 2022-11-27
Migrate cloud to astrolab

Herklos:
To be merged when cloud OctoBot is available on astrolab.cloud.

@GuillaumeDSM do you think we should first add a link to beta.astrolab.cloud?

### DIFF
--- a/.do/deploy.template.yaml
+++ b/.do/deploy.template.yaml
@@ -1,9 +1,0 @@
-spec:
-  name: octobot
-  services:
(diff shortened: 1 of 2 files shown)


GuillaumeDSM:
> To be merged when cloud OctoBot is available on astrolab.cloud.
> 
> @GuillaumeDSM do you think we should first add a link to beta.astrolab.cloud?

Yes, I think it would be great a add a section about astrolab beta telling that there is currently a closed beta where you can host your bot on

PR #2129, opened 2022-11-29
Env profiles

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Commons/pull/302


octobot/commands.py
```diff
@@ -150,6 +152,64 @@ async def install_all_tentacles(tentacles_url=None):
                                                               bot_install_dir=os.getcwd())
 
 
+def download_profile(url, target_file):
```
Herklos, line 155: Shouldn't we create an utility method to a download file?

octobot/commands.py
```diff
@@ -150,6 +152,64 @@ async def install_all_tentacles(tentacles_url=None):
...
+    # unauthenticated download
+    with requests.get(url, stream=True) as r:
+        r.raise_for_status()
+        with open(target_file, 'wb') as f:
+            for chunk in r.iter_content(chunk_size=8192):
+                f.write(chunk)
+    return target_file
+
+
+def _download_and_install_profile(download_url, logger):
```
Herklos, line 165: Why here and not in commons/profiles?

Herklos (approved):
:+1:

PR #2147, opened 2022-12-17
[SystemWatcher][ClockSynch] can disable from env variables

GuillaumeDSM:


Herklos:
> _ No description provided. _

What's the use case?

Herklos (approved):

PR #2149, opened 2022-12-22
Notify on dockerhub release

GuillaumeDSM:
image

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -201,3 +201,19 @@ jobs:
         with:
           status: Failure
           webhook: ${{ secrets.DISCORD_GITHUB_WEBHOOK }}
+
+
+  notify-dockerhub-update:
+    name: Notify dockerhub update
+    runs-on: ubuntu-latest
+    needs:
+      - lint
+      - build_test_push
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -201,3 +201,19 @@ jobs:
...
+    needs:
+      - lint
+      - build_test_push
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+
+    steps:
+      - name: Notify discord
+        uses: sarisia/actions-status-discord@v1
+        with:
+          description: "A new tag has been published on Docker Hub, docker bots can be updated."
```
Herklos, line 218: Should we tag @here ?

.github/workflows/docker.yml
```diff
@@ -201,3 +201,19 @@ jobs:
         with:
           status: Failure
           webhook: ${{ secrets.DISCORD_GITHUB_WEBHOOK }}
+
+
+  notify-dockerhub-update:
```
Herklos, line 206: 👍

.github/workflows/docker.yml
GuillaumeDSM: yes !

Herklos (approved):

PR #2180, opened 2023-01-11
Improve optimizer settings, live metadata and fix position size calculation

techfreaque:
requires https://github.com/Drakkar-Software/OctoBot-Trading/pull/797 for the `single_contract_value` position attribute

as discussed.
regarding your question:
- why those changes in optimizer settings ?
If one of the input settings is an empty string, it will set it to default values now


GuillaumeDSM:
Note: requires octobot_trading position udpated attributes

ruidazeng (approved):
LGTM

GuillaumeDSM:
to reopen if relevant

PR #2217, opened 2023-02-05
[Docker] add ntp to docker version

techfreaque:
requires PR in commons -> [ClockSync] add support for linux root users and docker


techfreaque:
Closed as ntp is not working in docker

PR #2223, opened 2023-02-10
[Optimizer] fix shuffle runs on runs less than select size

techfreaque:


octobot/strategy_optimizer/strategy_design_optimizer.py
```diff
@@ -733,7 +733,9 @@ async def _generate_and_store_backtesting_runs_schedule(self):
     def shuffle_and_select_runs(runs, select_size=None) -> dict:
         shuffled_runs = list(runs.values())
         random.shuffle(shuffled_runs)
-        selected_runs = shuffled_runs if select_size is None else shuffled_runs[:select_size]
+        selected_runs = (shuffled_runs 
+                         if (select_size is None or select_size <= len(shuffled_runs)) 
+                         else shuffled_runs[:select_size])
```
ruidazeng, line 738: The updated `shuffle_and_select_runs` function introduces a logic error, the if condition selects all runs when `select_size` is `None` or less than or equal to the number of runs, which means it ignores the `select_size` limit in cases where it should apply.

ruidazeng (changes requested):

octobot/strategy_optimizer/strategy_design_optimizer.py
techfreaque: Hi @ruidazeng 
oh I see, I changed it from <= to >=

ruidazeng (approved):

octobot/strategy_optimizer/strategy_design_optimizer.py
ruidazeng: @Herklos Is this something merge-able?

GuillaumeDSM:
to reopen if relevant

PR #2245, opened 2023-02-24
[Python3.10] upgrade to python 3.10

techfreaque:
Using it since a while without any issues. (restart and stop worked every single time, since we discussed it)

edit:
I changed this PR so it will only include the docker file changes and is on top of the other PR wich adds support for python 3.8, 3.9 and 3.10


Herklos:
Integrated in https://github.com/Drakkar-Software/OctoBot/pull/2327

PR #2246, opened 2023-02-24
[Optimizer] allow multi exchange backtests and handle settings fallbacks

techfreaque:
if the api receives a list of "exchange_ids" and/or a list of "data_files" it will optimize on multi exchanges.
Should behave the same way as before if "data_files" is a string and not a list and if no "exchange_ids" list is provided

requires:
https://github.com/Drakkar-Software/OctoBot-Trading/pull/847
https://github.com/Drakkar-Software/OctoBot-Backtesting/pull/172
https://github.com/Drakkar-Software/OctoBot-Tentacles/pull/865


octobot/strategy_optimizer/optimizer_settings.py
```diff
@@ -21,72 +21,142 @@
...
+        self.start_timestamp = settings_dict.get(
+            enums.OptimizerConfig.START_TIMESTAMP.value, None
+        )
+        self.end_timestamp = settings_dict.get(
+            enums.OptimizerConfig.END_TIMESTAMP.value, None
+        )
+        self.required_idle_cores = int(
+            settings_dict.get(enums.OptimizerConfig.IDLE_CORES.value)
+            or constants.OPTIMIZER_DEFAULT_REQUIRED_IDLE_CORES
+        )
```
ruidazeng, line 68: I think using `or` when assigning default values from `settings_dict` can cause incorrect defaults when valid falsy values (e.g., `0`, `False`) are provided. It might be better to change this to:
```python
self.required_idle_cores = int(
    settings_dict.get(
        enums.OptimizerConfig.IDLE_CORES.value,
        constants.OPTIMIZER_DEFAULT_REQUIRED_IDLE_CORES
    )
)
```

ruidazeng (changes requested):

octobot/strategy_optimizer/optimizer_settings.py
techfreaque: Hi @ruidazeng 
I see, I have removed those changes.

ruidazeng: @Herklos is this something merge-able? trying to backlog all the pull requests for OctoBot :))

GuillaumeDSM: @ruidazeng Thank you but we can see the open pull requests on the OctoBot repos, can you please not do this?

PR #2249, opened 2023-02-24
[Optimizer] properly handle multiple and single options

techfreaque:
You will need to adapt private optimizer code:
for ConfigTypes.MULTIPLE_OPTIONS to work properly (multiple-options user inputs), the strategy design config generated by the frontend needs to pass the config "type": ConfigTypes.MULTIPLE_OPTIONS.value  (config_element[self.CONFIG_TYPE] == ConfigTypes.MULTIPLE_OPTIONS.value)

all other ConfigTypes are supported to set via "type" as well, but should not be saved/sent via api as we can infer them still

### DIFF
--- a/octobot/strategy_optimizer/strategy_design_optimizer.py
+++ b/octobot/strategy_optimizer/strategy_design_optimizer.py
@@ -53,6 +53,7 @@ class ConfigTypes(enum.Enum):
     NUMBER = "number"
     BOOLEAN = "boolean"
     OPTIONS = "options"
+    MULTIPLE_OPTIONS = "multiple-options"
     UNKNOWN = "unknown"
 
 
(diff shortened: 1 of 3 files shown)


octobot/strategy_optimizer/strategy_design_optimizer.py
```diff
@@ -825,7 +832,26 @@ def _get_all_possible_values(self, start, stop, step):
...
+            
+    def _generate_possible_multiple_options_values(
+        self,
+        possible_values: list,
+        config_element_values,
+        parent_values: list = [],
+    ):
+        for combineable_value in config_element_values:
+            if combineable_value not in parent_values:
+                new_parent_values = parent_values + [combineable_value]
```
ruidazeng, line 844: Would it be better if we have something like this instead?

```python
def _generate_possible_multiple_options_values(
    self,
    possible_values: list,
    config_element_values,
    parent_values: Optional[list] = None,
):
    if parent_values is None:
        parent_values = []
    # rest of the code...
```

ruidazeng: Using a mutable default argument like `parent_values=[]` can lead to unintended side effects because the same list instance is shared across all calls to the function.

techfreaque: Hi @ruidazeng 
ah I see thanks, I changed it

ruidazeng: Awesome! @Herklos is this something merge-able?

JamesonDiamond1 (approved):
@JamesonDiamond1

GuillaumeDSM:
to reopen if relevant

PR #2299, opened 2023-03-29
Gpt req

GuillaumeDSM:


Herklos (approved):

octobot/backtesting/octobot_backtesting.py
```diff
@@ -235,6 +237,7 @@ def check_remaining_objects(self):
 
         for obj, max_ref in expected_max_objects_references.items():
             if objects_references[obj][0] > max_ref:
+                o = objects_references[obj][1][0]
```
Herklos, line 240: Is it used somewhere ?

GuillaumeDSM: 😮 
removing it

PR #2301, opened 2023-03-30
[Orders] add update_type to channel

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Trading/pull/884

### DIFF
--- a/octobot/logger.py
+++ b/octobot/logger.py
@@ -314,13 +314,13 @@ async def orders_callback(
         cryptocurrency: str,
         symbol: str,
         order: dict,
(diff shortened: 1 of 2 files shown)


octobot/logger.py
```diff
@@ -314,13 +314,14 @@ async def orders_callback(
         cryptocurrency: str,
         symbol: str,
         order: dict,
+        update_type: str,
```
Herklos, line 317: Should we remove is_new?

Herklos (approved):

PR #2325, opened 2023-05-02
python 3.10 update

GuillaumeDSM:


.github/workflows/docker.yml
```diff
@@ -49,7 +49,7 @@ jobs:
         id: wait-for-tentacles
         with:
           token: ${{ secrets.AUTH_TOKEN }}
-          checkName: "ubuntu-latest - Python 3.8 - upload tentacles"
+          checkName: "ubuntu-latest - Python 3.10 - upload tentacles"
```
Herklos, line 52: I think we will need to check if it's still the name after the migration

Herklos (approved):

Herklos (changes requested):

Herklos (approved):

PR #2449, opened 2023-10-22
[Community] handle candles history fetch

GuillaumeDSM:
requries https://github.com/Drakkar-Software/OctoBot-Trading/pull/993

### DIFF
--- a/octobot/community/supabase_backend/community_supabase_client.py
+++ b/octobot/community/supabase_backend/community_supabase_client.py
@@ -27,6 +27,8 @@ import supabase.lib.client_options
 import octobot_commons.authentication as authentication
 import octobot_commons.logging as commons_logging
 import octobot_commons.profiles as commons_profiles
+import octobot_commons.enums as commons_enums
+import octobot_commons.constants as commons_constants
 import octobot.constants as constants
 import octobot.community.errors as errors
 import octobot.community.supabase_backend.enums as enums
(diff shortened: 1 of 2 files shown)


octobot/community/supabase_backend/community_supabase_client.py
```diff
@@ -322,6 +324,91 @@ async def upsert_portfolio_history(self, portfolio_histories) -> list:
...
+        request_count = 0
+        while request_count < max_requests_count:
+            request = (
+                self.table("temp_ohlcv_history").select("*")
+                .eq(
+                    "exchange_internal_name", exchange
+                ).eq(
+                    "symbol", symbol
+                ).eq(
+                    "time_frame", time_frame.value
```
Herklos, line 365: You can use `match` instead of multiple `eq`

octobot/community/supabase_backend/community_supabase_client.py
```diff
@@ -322,6 +324,91 @@ async def upsert_portfolio_history(self, portfolio_histories) -> list:
...
+    ) -> list:
+        total_candles_count = (last_open_time - first_open_time) // (
+            commons_enums.TimeFramesMinutes[time_frame] * commons_constants.MINUTE_TO_SECONDS
+        )
+        offset = 0
+        max_size = 0
+        total_candles = []
+        max_requests_count = 100
+        request_count = 0
+        while request_count < max_requests_count:
```
Herklos, line 357: Why do we need to do multiple requests?

GuillaumeDSM: supabases has a max of rows returned by one query (1000 by default), here we want approx 18k rows for 15min over 6m

Herklos: And why not increasing supabase's limit?

GuillaumeDSM: according to supabase this value should not be too high as it can take a lot of computing power from the db and be an attack vector. We can increase it though, this code will just adapt

Herklos: Ok thanks for the explanation :+1:

octobot/community/supabase_backend/community_supabase_client.py
GuillaumeDSM: up, thanks !

octobot/community/supabase_backend/community_supabase_client.py
```diff
@@ -322,6 +324,89 @@ async def upsert_portfolio_history(self, portfolio_histories) -> list:
...
```
Herklos, line 357: Should we catch if a request fail to not stop the whole loop or it's ok to raise?

GuillaumeDSM: we would need a retry mechanism. I havn't seen it happen yet, do you think we should add it ?

Herklos: We can probably wait for it to happen a first time

Herklos (approved):

PR #2709, opened 2024-08-20
[TradingModes] allow starting without trading mode

GuillaumeDSM:


octobot/storage/trading_metadata.py
```diff
@@ -60,7 +60,11 @@ async def store_backtesting_run_metadata(exchange_managers, start_time, user_inp
 
 async def _get_trading_metadata(exchange_managers, run_start_time, user_inputs, run_dbs_identifier, is_backtesting, name) \
         -> dict:
-    trading_mode = trading_api.get_trading_modes(exchange_managers[0])[0]
+    try:
+        trading_mode = trading_api.get_trading_modes(exchange_managers[0])[0]
+    except IndexError:
+        import tentacles.Trading.Mode
+        trading_mode = tentacles.Trading.Mode.BlankTradingMode(exchange_managers[0].config, exchange_managers[0])
```
Herklos, line 67: I think we should log something when we fallback to BlankTradingMode

GuillaumeDSM: up

Herklos (approved):

PR #2712, opened 2024-08-21
[Sentry] add sentry tracker when dsn is set

GuillaumeDSM:


octobot/cli.py
```diff
@@ -362,43 +360,46 @@ def start_octobot(args):
         _disable_interface_from_param("web", args.no_web, logger)
 
         commands.run_bot(bot, logger)
-
+        force_error_exit = False
```
Herklos, line 363: 👍

Herklos (approved):
Nice feature !

PR #3172, opened 2025-12-19
[Community] fix auto re-auth on expired session error

GuillaumeDSM:


tests/unit_tests/community/test_authentication.py
```diff
@@ -414,6 +415,77 @@ def test_init_account(auth):
         assert auth._fetch_account_task == "task"
 
 
+async def test_bot_data_update(auth):
```
Herklos, line 418: 👌

PR #3212, opened 2026-02-03
[CI] Add and reuse build wheel step

Herklos:


.github/workflows/main.yml
```diff
@@ -9,7 +9,41 @@ on:
...
+  build:
+    name: ${{ matrix.os }}${{ matrix.arch }} - Python ${{ matrix.version }} - Build wheel
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ ubuntu-latest ]
+        arch: [ x64 ]
+        version: [ "3.13.x" ]
+    steps:
+      - uses: actions/checkout@v3
```
GuillaumeDSM:
```suggestion
        - uses: actions/checkout@v5
```
(I think we should use the same version everywhere

.github/workflows/main.yml
```diff
@@ -9,7 +9,41 @@ on:
...
+
+      - name: Set up Python 3.13
+        uses: actions/setup-python@v4
```
GuillaumeDSM:
```suggestion
        uses: actions/setup-python@v6
```
same here

Herklos:
Right, it's up

PR #3219, opened 2026-02-03
[Tests] parallelize tests when supported

GuillaumeDSM:


.github/workflows/main.yml
```diff
@@ -104,10 +104,13 @@ jobs:
       run: |
         if [ "${{ matrix.package }}" = "octobot" ]; then
           pytest tests
-          pytest --ignore=tentacles/Trading/Exchange tentacles
+          pytest --ignore=tentacles/Trading/Exchange tentacles -n auto --dist loadfile
         else
-          cd ${{ matrix.package }}
```
Herklos, line 109: we don't cd anymore?

GuillaumeDSM: woops

GuillaumeDSM:
it's up ! tests are now up to 2x faster

PR #3234, opened 2026-02-06
[Community] add community_bot and stop conditions

GuillaumeDSM:
Added:
- community bot management in octobot/community/community_bot.py to handle bot logs, bot error status etc
- stop conditions mecanisms as automations
- bot stop method in octobot to stop trading on all exchanges


octobot/producers/exchange_producer.py
Herklos: I think it should use octobot_channel_callback with probably a new OctoBotChannelSubjects.STOP.value to better match the architecture (to use the same way as creation)

octobot/octobot_api.py
```diff
@@ -13,16 +13,35 @@
...
+        self.octobot_api_by_bot_id: dict[str, "OctoBotAPI"] = {}
+
+    def get_api(self, bot_id: str) -> "OctoBotAPI":
+        return self.octobot_api_by_bot_id[bot_id]
+
+    def register_api(self, bot_id: str, octobot_api: "OctoBotAPI"):
+        self.octobot_api_by_bot_id[bot_id] = octobot_api
+
+    def unregister_api(self, bot_id: str):
+        self.octobot_api_by_bot_id.pop(bot_id)
```
Herklos, line 37: Do we really need this? I think we already have something equivalent to get the octobot API instance

packages/tentacles/Automation/trigger_events/holding_threshold_event/holding_threshold.py
```diff
@@ -0,0 +1,148 @@
...
+        self.asset_name = config[self.ASSET_NAME]
+        self.amount = decimal.Decimal(str(config[self.AMOUNT]))
+        self.stop_on_inferior = config[self.STOP_ON_INFERIOR]
+        if not self.exchange or not self.asset_name or not self.amount:
+            raise errors.InvalidAutomationConfigError("Exchange, asset name and amount must be set", self.get_name())
+
+    def _check_threshold(
+        self, exchange_manager: "trading_exchanges.ExchangeManager"
+    ) -> tuple[bool, typing.Optional[str]]:
+        holdings = exchange_manager.exchange_personal_data.portfolio_manager.portfolio.get_currency_portfolio(self.asset_name)
```
Herklos, line 101: I think it should use the trading api

packages/tentacles/Automation/trigger_events/holding_threshold_event/holding_threshold.py
```diff
@@ -0,0 +1,148 @@
...
+        return (
+            f"Current {self.asset_name} holdings of {float(holdings.total)} are "
+            f"{'lower' if self.stop_on_inferior else 'higher'} than the {float(self.amount)} threshold."
+        )
+
+    async def register_consumer(self):
+        await self._register_exchange_channel_consumer(
+            channel_name=channels_name.OctoBotTradingChannelsName.BALANCE_CHANNEL.value,
+            exchange=self.exchange,
+        )
```
Herklos: 👍

packages/trading/octobot_trading/exchanges/traders/trader.py
```diff
@@ -55,6 +56,19 @@ def enabled_or_forced_only_wrapper(self, *args, force_if_disabled=False, **kwarg
...
+    """
+    Decorator to call on_invalid_credentials if an AuthenticationError is raised.
+    """
+    async def authenticated_wrapper(self, *args, **kwargs):
+        try:
+            return await func(self, *args, **kwargs)
+        except errors.AuthenticationError as err:
+            await self.on_invalid_credentials()
+            raise err
+    return authenticated_wrapper
```
Herklos, line 69: 👍

octobot/automation/bases/abstract_action.py
```diff
@@ -16,8 +16,19 @@
...
 
 class AbstractAction(automation_step.AutomationStep, abc.ABC):
-    async def process(self):
+    async def call_process(
+        self, execution_details: execution_details.ExecutionDetails
+    ) -> bool:
+        if await self.process(execution_details):
+            self.update_last_execution_details(execution_details)
+            return True
+        return False
```
Herklos: Shouldn't we use a decorator on process instead?

octobot/automation/bases/abstract_channel_based_trigger_event.py
```diff
@@ -0,0 +1,126 @@
...
+            filter_kwargs["symbol"] = symbol
+        self._consumers.append(
+            await exchanges_channel.get_chan(
+                channel_name,  trading_api.get_exchange_manager_id(exchange_manager)
+            ).new_consumer(
+                self.channel_callback,
+                priority_level=channel_enums.ChannelConsumerPriorityLevels.HIGH.value,
+                **filter_kwargs
+            )
+        )
```
Herklos: I think the caller should define a dedicated callback function instead of a generic one

octobot/octobot.py
```diff
@@ -204,6 +206,25 @@ async def _store_run_metadata_when_available(self):
...
+        except Exception as err:
+            self.logger.exception(err, True, f"Error when stopping trading modes: {err}")
+        try:
+            await self.community_auth.community_bot.schedule_bot_stop(stop_reason)
+        except Exception as err:
+            self.logger.exception(err, True, f"Error when scheduling bot stop: {err}")
+        if execution_details is not None:
+            await self.community_auth.community_bot.insert_stopped_strategy_execution_log(
+                execution_details.description
+            )
```
Herklos: I think this function should not be here but should only be in octobot_api. There is no need of this function here

octobot/automation/bases/abstract_action.py
GuillaumeDSM: if we decorate process, we will have to use this decorator in all subclasses as process is overriden there. I added a decodator on call_process instead

octobot/producers/exchange_producer.py
GuillaumeDSM: we can do this however if we do we won't be able to use it from backtesting / non "full" octobots as they don't use octobot channels for their init

Herklos: I don't understand your point. As backtesting doesn't use octobot channels, why do we put this code in octobot/producers ? It doesn't make sense if you want to be able to use this code in a context that doesn't use octobot channels

PR #3237, opened 2026-02-11
[ProfileCopyTrading] Fix ref market ratio calculation

Herklos:
Related to https://github.com/Drakkar-Software/OctoBot-Prediction-Market/issues/3

### DIFF
--- a/packages/tentacles/Trading/Mode/profile_copy_trading_mode/profile_copy_trading.py
+++ b/packages/tentacles/Trading/Mode/profile_copy_trading_mode/profile_copy_trading.py
@@ -216,7 +216,13 @@ class ProfileCopyTradingMode(index_trading_mode.IndexTradingMode):
         self.total_ratio_per_asset = global_distribution[profile_distribution.TOTAL_RATIO_PER_ASSET]
         self.indexed_coins = global_distribution[profile_distribution.INDEXED_COINS]
         self.indexed_coins_prices = global_distribution[profile_distribution.INDEXED_COINS_PRICES]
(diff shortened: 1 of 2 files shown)


packages/tentacles/Trading/Mode/profile_copy_trading_mode/profile_copy_trading.py
```diff
@@ -216,7 +216,13 @@ def update_global_distribution(self):
         self.total_ratio_per_asset = global_distribution[profile_distribution.TOTAL_RATIO_PER_ASSET]
         self.indexed_coins = global_distribution[profile_distribution.INDEXED_COINS]
         self.indexed_coins_prices = global_distribution[profile_distribution.INDEXED_COINS_PRICES]
-        self.reference_market_ratio = global_distribution[profile_distribution.REFERENCE_MARKET_RATIO]
+        # IndexTradingMode consumes reference_market_ratio as the tradable portfolio ratio.
+        # Profile distribution returns the reserved reference-market ratio, so convert it.
+        reference_market_reserved_ratio = global_distribution[profile_distribution.REFERENCE_MARKET_RATIO]
```
GuillaumeDSM, line 221: REFERENCE_MARKET_RATIO in config is what we don't want to trade ? I thought we changed it to what we want to trade

Herklos: We are in profile copy trading mode here, not in index trading mode. In index trading mode it comes from config, in profile it comes from the distribution calculation

Herklos: I forgot to update this code after the change we made on config

PR #3241, opened 2026-02-12
[Stop] fix multiple calls and add pause only option

GuillaumeDSM:


octobot/community/community_bot.py
```diff
@@ -90,19 +91,45 @@ async def on_started_bot(self):
...
+    ):
+        with caught_global_exceptions(
+            f"on_trading_modes_stopped_and_traders_paused: {stop_reason=}"
+        ):
+            if schedule_bot_stop:
+                # schedule bot stop: bot will pause trading and be stopped
+                await self.schedule_bot_stop(stop_reason)
+            elif stop_reason is not None:
+                # only update error status: bot will pause trading but remain on
+                await self.update_deployment_error_status_for_stop_reason(stop_reason)
```
Herklos, line 112: 👍

octobot/producers/exchange_producer.py
```diff
@@ -73,6 +76,7 @@ async def stop(self):
         self.logger.debug("Stopping ...")
         for exchange_manager in trading_api.get_exchange_managers_from_exchange_ids(self.exchange_manager_ids):
             await trading_api.stop_exchange(exchange_manager)
+        self.stopped_trading_modes_and_traders = True
```
Herklos: Shouldn't we get the values from trading_api that would check the real value of traders and trading mode. I'm concerned about a possible desynchronization here that could be avoided by checking the real value.

packages/commons/octobot_commons/cache_util.py
```diff
@@ -0,0 +1,36 @@
...
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+#  Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser General Public
+#  License along with this library.
+import time
+
+
+def prevented_multiple_calls(f):
```
Herklos, line 19: 👍

Herklos (approved):

PR #3245, opened 2026-02-12
[IndexTradingMode] Fix futures trading selling coins and orders cance…

Herklos:


packages/tentacles/Trading/Mode/index_trading_mode/rebalancer/rebalancer.py
```diff
@@ -75,6 +75,17 @@ async def get_coins_to_sell_orders(self, details: dict, dependencies: typing.Opt
             self.trading_mode.exchange_manager.exchange_personal_data.portfolio_manager.reference_market, {},
             dependencies=dependencies
         )
+        # When a coin has no sellable holdings, conversion won't emit a sell order.
+        # In that case, cleanup stale open SELL orders for this symbol to avoid endless sell attempts.
+        portfolio = self.trading_mode.exchange_manager.exchange_personal_data.portfolio_manager.portfolio
+        for coin in order_coins_to_sell:
+            symbol, base_asset = self.get_symbol_and_base_asset(coin)
+            if portfolio.get_currency_portfolio(base_asset).total <= decimal.Decimal(0):
```
GuillaumeDSM: I don't understand this condition, why are we checking if holdings are negative ?
I guess we eeither want to check if there are open orders or if available<total (which means some of it is locked in orders)

Herklos: `<=` is probably overkill, we may prefer `==`. The idea here is to check if we have holdings or we only have open orders. If we only have open orders we should only cancel them

packages/tentacles/Trading/Mode/index_trading_mode/tests/test_index_trading_mode.py
```diff
@@ -1945,74 +1945,474 @@ async def test_get_simple_buy_coins(trading_tools):
...
+            mode, "cancel_order", mock.AsyncMock(return_value=(True, commons_signals.SignalDependencies()))
+        ) as cancel_order_mock, mock.patch.object(
+            trading_personal_data, "wait_for_order_fill", mock.AsyncMock()
+        ):
+            with pytest.raises(trading_errors.MissingMinimalExchangeTradeVolume):
+                await consumer.trading_mode.rebalancer.sell_indexed_coins_for_reference_market(details, dependencies)
+        return cancel_order_mock
+
+    details = scenario["details"]
+    if scenario["flow"] == "buy":
```
GuillaumeDSM: considering the sell and buy flows are completely different branches in this test, with almost nothing in common (except 3 init lines), can we create 2 tests instead of this very complex pytest.parametrize?

Herklos: Right, I'll do that

PR #3268, opened 2026-02-23
[CopyTrading] Fix futures holdings ration calculation after buy order fill

Herklos:


packages/trading/tests/personal_data/portfolios/test_portfolio_value_holder.py
```diff
@@ -510,6 +514,44 @@ def mock_get_contract(pair):
             "BTC", traded_symbols_only=True, include_assets_in_open_orders=True
         ) == decimal.Decimal('0.9166666666666666666666666667')
 
+        # Test inflated ratio when a buy order is open but its fill has not yet been confirmed
+        if exchange_manager.is_future or exchange_manager.is_option:
+            order_quantity = decimal.Decimal("100")
+            order_price = decimal.Decimal("0.3")
+            pending_order_value = order_quantity * order_price  # = 30 USDC
+            portfolio_value_holder.portfolio_current_value = decimal.Decimal("70")
```
GuillaumeDSM, line 522: in futures the portfolio current value excludes funds that are locked in buy order ? 😮 Or I missunderstand

Herklos: It's true on polymarket. If other futures exchange don't work this way, I would need to add a specific code for exchanges like polymarket somewhere else

PR #3269, opened 2026-02-23
[Commons] Add monitored process

Herklos:


packages/commons/octobot_commons/monitored_process.py
```diff
@@ -0,0 +1,370 @@
...
+                self.__class__.__name__,
+                'gracefully' if graceful else 'forcibly',
+            )
+        self._process = None
+
+        self._stderr_buffer.clear()
+        self._stdout_buffer.clear()
+
+        if self._monitor_error is not None:
+            raise self._monitor_error
```
GuillaumeDSM, line 282: I think returning something in aexist can suppress exceptions, can we add a test to be sure it's not the case here and if it is, remore this return?

packages/commons/octobot_commons/monitored_process.py
```diff
@@ -0,0 +1,370 @@
...
+    def __init__(self, exit_code: typing.Optional[int], stderr: typing.Optional[str] = None):
+        self.exit_code = exit_code
+        self.stderr = stderr
+        msg = f"Process exited prematurely with code {exit_code}"
+        if stderr:
+            msg += f"; stderr: {stderr}"
+        super().__init__(msg)
+
+
+class MonitoredProcess:  # pylint: disable=too-many-instance-attributes
```
GuillaumeDSM, line 113: 👍

packages/commons/octobot_commons/monitored_process.py
Herklos: PR is ready with a new test for these lines

PR #3290, opened 2026-03-04
[POlymarket] Fix recent issues

Herklos:


packages/tentacles/Trading/Mode/index_trading_mode/tests/test_index_trading_mode.py
```diff
@@ -2168,6 +2170,112 @@ async def test_sell_some_reduces_or_closes_position(trading_tools):
             assert orders == []
 
 
+@pytest.mark.parametrize("trading_tools", ["futures"], indirect=True)
```
GuillaumeDSM: we should remove those @pytest.mark.parametrize when there is only 1 value to use

Herklos: Right! It's up

PR #3322, opened 2026-03-21
[GPTService] Restore config UI

GuillaumeDSM:
image

- restore GPT config from UI
- improve custom URL connection error
- renamed "base-url" to "llm-custom-base-url" to align with previous config key (I think the previous value is still valid)
- renamed "auth-token" into "api-key" with retrocompatibility with previous config (key was openai-secret-key)


packages/tentacles/Services/Services_bases/gpt_service/gpt.py
```diff
@@ -126,25 +122,28 @@ def get_fields_description(self):
...
-                services_constants.CONFIG_LLM_MCP_SERVERS: [],
-                services_constants.CONFIG_LLM_AUTO_INJECT_MCP_TOOLS: True,
-                services_constants.CONFIG_LLM_TOOL_CALL_JSON_OUTPUT: True,
-                services_constants.CONFIG_LLM_AI_PROVIDER: "",
-                services_constants.CONFIG_LLM_AUTH_TOKEN: "",
+                # return config values to use in UI
+                services_constants.CONFIG_LLM_API_KEY: "",
+                services_constants.CONFIG_LLM_CUSTOM_BASE_URL: "",
+                # other config values are available but not displayed in UI for now
+                # services_constants.CONFIG_LLM_MODEL: self.DEFAULT_MODEL,
```
Herklos: Shouldn't the model be configured?

Herklos (commented):
Thanks for these fixes 👍

packages/tentacles/Services/Services_bases/gpt_service/gpt.py
GuillaumeDSM: When using the GPTEvaluator, it's the evaluator that defines the model to use
image

For now I see no need to define it from the UI but I might be wrong

Herklos: I see, GPTEvaluator is not the only way we use this Service anymore (AiStrategy, AITradingMode...). I think we should define a default model in LLMService so that if not defined in the evaluator, strategy, tradingmode or whatever, it uses the default model

GuillaumeDSM: ok, i'll add it

GuillaumeDSM: image

new selector option in GPTEvaluator config to align to the GPT interface value


=> model added to the GPT interface config
image

Herklos (approved):

PR #3344, opened 2026-03-24
[Node] Improve UX and cleanup

Herklos:


packages/node/octobot_node/scheduler/api.py
```diff
@@ -134,6 +134,21 @@ async def get_all_tasks() -> list[octobot_node.models.Task]:
     return tasks
 
 
+async def delete_task(task_id: str) -> str:
+    all_workflows = await octobot_node.scheduler.SCHEDULER.INSTANCE.list_workflows_async(status=[
+        "ENQUEUED", "PENDING", "SUCCESS", "ERROR", "CANCELLED", "MAX_RECOVERY_ATTEMPTS_EXCEEDED"
```
GuillaumeDSM: we should use `dbos.WorkflowStatusString` for status instead of hard coded strings when possible

packages/node/octobot_node/scheduler/api.py
```diff
@@ -134,6 +134,21 @@ async def get_all_tasks() -> list[octobot_node.models.Task]:
...
+    ])
+    parent_id = task_id[:octobot_node.constants.PARENT_WORKFLOW_ID_LENGTH]
+    matching_workflow_ids = [
+        wf.workflow_id for wf in all_workflows
+        if wf.workflow_id[:octobot_node.constants.PARENT_WORKFLOW_ID_LENGTH] == parent_id
+    ]
```
GuillaumeDSM: this is already done in schheduler.delete_workflows why do we do it again?

Herklos: Right, I didn't check the scheduler implementation

Herklos:
It's up thanks

packages/node/octobot_node/scheduler/api.py
```diff
@@ -134,6 +140,11 @@ async def get_all_tasks() -> list[octobot_node.models.Task]:
     return tasks
 
 
+async def delete_task(task_id: str) -> str:
+    await octobot_node.scheduler.SCHEDULER.delete_workflows([task_id])
```
GuillaumeDSM, line 144: 👍

GuillaumeDSM (approved):

PR #3364, opened 2026-04-02
[Node] smart CSV import/export with templates

Herklos:


packages/node/octobot_node/scheduler/api.py
```diff
@@ -109,11 +109,14 @@ def _build_tasks_from_executions(
     tasks = []
     for parent_id, group in grouped.items():
         active = _get_active_execution(group)
+        error = next((e.error for e in group if e.error), None)
```
GuillaumeDSM: consierding the DSL stops at the first error, the "active" execution should contain the error we want to display I guess. why taking the first (encountered error instead?)

packages/tentacles/Services/Interfaces/node_web_interface/package.json
```diff
@@ -8,7 +8,9 @@
     "build": "tsc -p tsconfig.build.json && vite build",
     "lint": "biome check --write --unsafe --no-errors-on-unmatched --files-ignore-unknown=true ./",
     "preview": "vite preview",
-    "generate-client": "openapi-ts"
+    "generate-client": "openapi-ts",
+    "test": "vitest run",
```
GuillaumeDSM, line 12: should we update something in CI to executee those tests ?

GuillaumeDSM (commented):
great work !!

Herklos:
Right thanks! it's up

GuillaumeDSM (approved):
🎉

PR #3439, opened 2026-05-04
[Sync] use encode_defunct for EIP-191 signing to avoid double

Herklos:


packages/sync/octobot_sync/auth/canonical.py
```diff
@@ -15,22 +15,12 @@
 #  License along with OctoBot. If not, see <https://www.gnu.org/licenses/>.
 
 import hashlib
-import typing
 
+from eth_account.messages import encode_defunct, SignableMessage
```
GuillaumeDSM: I think you need to pip uninstall eth_account 😂

Herklos: It's up

GuillaumeDSM (approved):
👍

PR #3469, opened 2026-05-21
[Node] improve error message and add retry attempts

GuillaumeDSM:


packages/tentacles/Services/Interfaces/node_web_interface/src/utils/task-status.ts
```diff
@@ -45,7 +46,7 @@ export function getStatusVariant(
 export function getTaskFilterGroup(task: Task): TaskFilterGroup {
   const exec = getActiveExecution(task.executions)
   if (getStatusGroup(exec?.status) === "active") return "active"
-  return exec?.status === "failed" || exec?.error ? "errored" : "completed"
+  return hasTaskError(task) ? "errored" : "completed"
```
Herklos: I don't think we want to change this check

Herklos: I don't think we should use `hasTaskError` at all

GuillaumeDSM: thanks, it's up!

Herklos (approved):

PR #3487, opened 2026-06-10
[CI] add npm trusted publishing for octobot-protocol

Herklos:


.github/workflows/main.yml
```diff
@@ -504,6 +504,38 @@ jobs:
           workingDirectory: docs
           command: versions deploy --yes
 
+  publish-npm:
+    name: Publish npm packages
+    needs: [ build, tests ]
+    if: github.event_name == 'push' && (github.ref == 'refs/heads/master' || github.ref == 'refs/heads/dev')
```
GuillaumeDSM, line 510: do we need a tag condition here or we publish at every dev/master push?

Herklos: As specified in the step name, we check if the package version exists on npm, if it doesn't exist we publish the version

GuillaumeDSM (approved):

PR #3502, opened 2026-06-16
[Copy] early abord grace period on no open order or market rebalance

GuillaumeDSM:


packages/flow/octobot_flow/parsers/actions_dag_parser.py
```diff
@@ -382,6 +382,7 @@ def parse(self) -> octobot_flow.entities.ActionsDAG:
         )
         actions_dag = octobot_flow.entities.ActionsDAG([init_action])
         self._parse_generic_actions(actions_dag)
+        print("dag", actions_dag)
```
Herklos: Should we keep this log ?

GuillaumeDSM: right, it's up, thanks!

Herklos (approved):

PR #3512, opened 2026-06-24
[Node] handle full bot start

GuillaumeDSM:
fixes many bugs related to process bots, now everything seens to work fine (and is tested)


packages/tentacles/Services/Interfaces/node_web_interface/src/lib/debug/user-action-templates.ts
```diff
@@ -318,6 +323,19 @@ function sampleGenericProcessStrategyConfiguration(
...
+function sampleGenericProcessOctobotStrategyConfiguration(
+  id = "<strategy-id>",
+): Strategy {
+  return sampleStrategyShell(
+    id,
+    "Generic process OctoBot strategy",
+    {
+      configuration_type: "generic_process",
+    } satisfies GenericProcessConfiguration,
+    "USDC",
```
Herklos, line 335: Shouldn't the ref market be a param too alongside id?

Herklos (commented):
👍

packages/tentacles/Services/Interfaces/node_web_interface/src/lib/debug/user-action-templates.ts
GuillaumeDSM: it could but it's just a template here, we just want default values. id is a param to generate a new random id at each call but we don't have this need for ref markets

Herklos (approved):

PR #3513, opened 2026-06-25
[Node] fix process bot restart issues

GuillaumeDSM:
there was 2 issues:
1. we need to handle process octobot restarts (ex: when changing config) => this changes pid, node has to bind to the new pid and not start a duplicate process (also needs a grace period to handle the few secs when bot is starting and did not yet update its state file)
2. when master is stopped, it stops child processes, when it's restarted, we want to instantly restart the child without waiting for any grace period (child is not "late", it's just hasnt been respawned yet)

### DIFF
--- a/packages/node/octobot_node/constants.py
+++ b/packages/node/octobot_node/constants.py
@@ -14,6 +14,7 @@
 #  You should have received a copy of the GNU General Public
 #  License along with OctoBot. If not, see <https://www.gnu.org/licenses/>.
 import os
+import uuid
 
 try:
     import octobot.constants as octobot_constants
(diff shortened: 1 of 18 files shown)


packages/tentacles/Services/Interfaces/node_web_interface/metadata.json
```diff
@@ -3,6 +3,6 @@
   "origin_package": "OctoBot-Default-Tentacles",
   "tentacles": ["NodeWebInterface"],
   "tentacles-requirements": ["node_api_interface"],
-  "build": ["npm install", "npm run generate-client", "npm run build"],
```
Herklos, line 6: 😮

GuillaumeDSM: oups

packages/node/octobot_node/constants.py
```diff
@@ -67,3 +68,5 @@
 )
 
 DEFAULT_PORTFOLIO_VALUATION_UNIT = "USDT"
+
+SCHEDULER_EXECUTOR_ID = str(uuid.uuid4()) # unique for each worker
```
Herklos, line 72: 👍

Herklos (approved):

Herklos (approved):

PR #3526, opened 2026-07-09
Fixes and more

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot/pull/3520

PR #3530, opened 2026-07-12
Node by default

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot/pull/3526

1. use node mode by default
2. display process bot names in their UI
3. show link to process bots in nodes automations cards


Herklos (approved):
👍

Issue #3597, opened 2026-07-31
Error when updating/setting hostname: AttributeError: 'NoneType' object has no attribute 'config'


blueyed:
When trying to set a hostname in the settings I am getting an internal server error:

> octobot-1  |   File "/opt/venv/lib/python3.13/site-packages/octobot_services/services/abstract_service.py", line 178, in save_service_config                                                                         
octobot-1  |     if update and service_key in self.edited_config.config[constants.CONFIG_CATEGORY_SERVICES]:                                                                                                         
octobot-1  |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^                                   
octobot-1  | AttributeError: 'NoneType' object has no attribute 'config'                                  

Version 3.0.0-beta1.

GuillaumeDSM:
I can't reproduce your issue (fresh install editing domain from settings works on my test). can you please give more info on how to reproduce it ?

blueyed:
On `/app/settings` I add my hostname (foo.bar.com), and click Save.

Let me know if you need the full traceback, and some print-debugging.

GuillaumeDSM:
It's excatly what I did I seems I have no error. Can you give me more context please ? like is it a new install, is it on docker, is it on a local or remote server?

blueyed:
Running in Docker, upgraded using the button, later pulled the latest image.

I've upgraded from 2.1.1.

A bit more context:

> octobot-1  |   File "/octobot/tentacles/Services/Interfaces/node_api_interface/api/routes/nodes.py", line 104, in update_node_config
octobot-1  |     node_api_service_module.NodeApiService.instance().set_node_external_host(config.external_host)
octobot-1  |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
octobot-1  |   File "/octobot/tentacles/Services/Services_bases/node_api_service/node_api.py", line 203, in set_node_external_host
octobot-1  |     self.save_service_config(
octobot-1  |     ~~~~~~~~~~~~~~~~~~~~~~~~^
octobot-1  |         services_constants.CONFIG_NODE_API,
octobot-1  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
octobot-1  |         {services_constants.NODE_EXTERNAL_HOST: self.node_external_host},
[...]

PR #3599, opened 2026-07-31
[Node] fix AttributeError when saving node config

Herklos:
Related to https://github.com/Drakkar-Software/OctoBot/issues/3597


GuillaumeDSM (approved):
👍

PR #3621, opened 2026-08-05
[Node] never store keys in exported collection models

GuillaumeDSM:


packages/node/octobot_node/protocol/util/privacy_filter.py
```diff
@@ -16,135 +16,108 @@
...
+})
 
-_ACCOUNT_AUTHENTICATION_SENSITIVE_FIELD_NAMES = (
+_ACCOUNT_AUTHENTICATION_SENSITIVE_FIELD_NAMES = frozenset({
     "api_key",
     "api_secret",
     "api_passphrase",
     "public_key",
     "private_key",
     "seed_phrase",
```
Herklos: Couldn't we use trading constants directly so that we are sure we use the correct value?

GuillaumeDSM: right, it's up

Herklos (approved):

Issue #3611, opened 2026-08-02
GridTradingMode creates single order(s) only


blueyed:
In version 3.0.0-beta1 the grid trading mode creates only a single buy and sell order, although it is configured to have 5.

The amount is 1/5, which is correct.  But there should be 4 more orders on every side.

This results in a warning:

> WARNING	GridTradingModeProducer	SOL/USD1 trailing up process aborted: too many missing buy orders. Only 1 are online while configured total orders is 10

The "Create initital orders" action does nothing.

blueyed:
This works in 3.0.0-beta2 now.

blueyed:
Still an issue, although it worked as expected once.

Herklos:
> Still an issue, although it worked as expected once.

Hey @blueyed, thanks for reporting. It's the same issue or another one?

blueyed:
The same.  Not really clear why it worked or what is failing exactly though.

Herklos:
Can you please open a ticket in your octobot node settings, then go back to octobot node home, select your grid bot and click "share logs"? It will share anonymous logs with us so that we can have more debug information on our side.

blueyed:
What I've just seen is this 5 times:

> GridTradingModeConsumer	No order created for OrderData(side=<TradeOrderSide.BUY: 'buy'>, quantity=Decimal('0.01601447190635630337313481511'), price=Decimal('75.269'), symbol='SOL/USD1', is_virtual=False, associated_entry_id=None) (cost: 1.205393285919532598592484399): incompatible with exchange minimum rules. Limits: {'leverage': {'min': None, 'max': None}, 'amount': {'min': 0.0, 'max': None}, 'price': {'min': None, 'max': None}, 'cost': {'min': 1.0, 'max': 2000000.0}}

In the app I can order either 0.01 or 0.02, but not 0.016, which might be the issue here.

The total amount in the portfolio is 6.033 USD1.  It is configured to create 5 buy orders.

blueyed:
> Can you please open a ticket in your octobot node settings, then go back to octobot node home, select your grid bot and click "share logs"? It will share anonymous logs with us so that we can have more debug information on our side.

Done. But I can only share logs when the ticket got accepted..?!

blueyed:
Here is a warning, where the limits should fit.  It was created after a BUY was triggered.

[...]

PR #3648, opened 2026-08-15
staggered_orders_trading: use order_quantity/order_price in warning

blueyed:
Ref: https://github.com/Drakkar-Software/OctoBot/issues/3611#issuecomment-5302151504

### DIFF
--- a/packages/tentacles/Trading/Mode/staggered_orders_trading_mode/staggered_orders_trading.py
+++ b/packages/tentacles/Trading/Mode/staggered_orders_trading_mode/staggered_orders_trading.py
@@ -544,7 +544,7 @@ class StaggeredOrdersTradingModeConsumer(trading_modes.AbstractTradingModeConsum
                 )
             if not created_order:
                 self.logger.warning(
-                    f"No order created for {order_data} (cost: {quantity * order_data.price}): "
+                    f"No order created for {order_data} (cost: {order_quantity * order_price}): "
                     f"incompatible with exchange minimum rules. "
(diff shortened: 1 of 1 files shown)


packages/tentacles/Trading/Mode/staggered_orders_trading_mode/staggered_orders_trading.py
```diff
@@ -544,7 +544,7 @@ async def create_order(
                 )
             if not created_order:
                 self.logger.warning(
-                    f"No order created for {order_data} (cost: {quantity * order_data.price}): "
+                    f"No order created for {order_data} (cost: {order_quantity * order_price}): "
```
GuillaumeDSM, line 547: order_quantity and order_price are initialized in the above for loop that can yield no iteration, in this case, order_quantity and order_price would be unset. To use them in this log, we need to init them before the loop, to be sure they are always set variables

blueyed: Ok.
Can you let the workflow/tests run? I'm interested to see if there are falling tests.

GuillaumeDSM: tests are not failing atm, it's just a risk if we include this change.
You should be able to run tests on your own setup using pytest with the proper PYTHON_PATH and giving this as argument `tentacles/Trading/Mode/staggered_orders_trading_mode` (to run staggered_orders_trading_mode tests only)

Issue #3647, opened 2026-08-15
Memory leak on /app/octobots


blueyed:
When accessing the list of octobots on the node, the Octobot process starts to eat up all available memory (3GB RAM + 2GB swap), and I have to restart the Docker container to fix the system.

There are two octobots being shown, but I have created 3.

Version 3.0.0-beta2.

blueyed:
This happens while loading the octobots.

blueyed:
@Herklos thanks for addressing this. Is there a workaround to create a new Octobot?

PR #3653, opened 2026-08-23
[Node] fix /app/octobots browser memory leak

Herklos:
Try to fix https://github.com/Drakkar-Software/OctoBot/issues/3647


GuillaumeDSM (approved):
👍

PR #3658, opened 2026-08-24
[Protocol] Add more types

Herklos:


packages/tentacles/Services/Interfaces/node_web_interface/package.json
```diff
@@ -13,6 +13,7 @@
     "test:watch": "vitest"
   },
   "dependencies": {
+    "@drakkar.software/octobot-client": "file:../../../../client/octobot_client_ts",
```
GuillaumeDSM, line 19: should we keep this?

packages/protocol/openapi.json
```diff
@@ -918,6 +940,66 @@
           }
         }
       },
+      "BrokerAccount": {
+        "description": "BrokerAccount",
```
GuillaumeDSM: just to be sure we don't forget it: could we add a description to all AccountTypes so that we can't mix them?

packages/tentacles/Services/Interfaces/node_web_interface/package.json
Herklos: Yes, matching the same pattern `docs/package.json` already uses for this exact package (also `file:` linked, also not published yet). What was actually missing was the build-ordering piece: this package's own `npm run build` never built the sibling `octobot_client_ts` first, and its `dist/` is gitignored, so a fresh checkout failed. Added `sdk:build`/`prebuild`/`pretest` scripts mirroring docs' own `sdk:build` hook, so npm's lifecycle runs it automatically before `build`/`test` regardless of caller (including the tentacle exporter's own `npm run build` call).

One caveat worth flagging: this makes the mechanism correct, but `packages/copy`'s CI job still can't go fully green on this branch today, because `octobot_client_ts` itself depends on `octobot-protocol@^0.8.0`, which isn't published on the real npm registry yet (only up to 0.6.0 is). Same root cause as the client package not being published past 0.5.0. It'll resolve once this PR merges and the real publish runs. `packages/copy` isn't a required status check, so it doesn't block merging in the meantime.

packages/protocol/openapi.json
Herklos: Added — `AccountType`'s description now spells out what distinguishes all six values (each already has the correct $ref+description convention on its own `account_type` property, so narrowing itself was never at risk; this was purely missing docs):

> Account category discriminator. `generic`: a placeholder account created before its real kind is known (e.g. mid-onboarding), not a semantic category of its own. `exchange`: a centralized or DEX trading account, live-synced via a provider connection. `blockchain`: a wallet address tracked read-only on-chain. `broker`: a brokerage account for trading stocks, commodities, forex, and similar instruments (eToro-style platforms), which may be live-synced via a provider connection. `bank`: a cash/deposit account, always manually entered, never live-synced. `asset`: a manually-entered, single tracked non-liquid asset (a watch, a car, real estate, and similar).

Regenerated via `npm run generate:all` (129/129 Python model tests green, no shape change).

Herklos:
it's up

packages/protocol/openapi.json
GuillaumeDSM: can we put the definition of each account type in the relevant account type and not refer other account types in the generic one ?

Herklos: Right, sorry, it's up !

packages/tentacles/Services/Interfaces/node_web_interface/package.json
Herklos: It needs the octobot PR to be merged first

GuillaumeDSM (approved):
👍

PR #3667, opened 2026-09-02
[Protocol] add compatibility tests framework

GuillaumeDSM:
on protocol:
- make sure code is sync with openapi.json
- fails on unhandled breaking change
- ensure previous formats are never breaking the latest format

on sync & node & flow:
- make sure the latest version of protocol properly loads

on sync: also load all previous breaking versions to be use they load as well 

everything runs automatically alongside other tests


packages/protocol/scripts/build_openapi_schema_manifest.py
```diff
@@ -0,0 +1,36 @@
+#  Drakkar-Software OctoBot-Protocol
+#  Copyright (c) 2026 Drakkar-Software, All rights reserved.
+#
+# CLI: regenerate test/compat/static/openapi_schema_manifest.json from openapi.json.
+#
+# Run after editing openapi.json (non-breaking: same PR; breaking: after migration work):
+#   python -m scripts.build_openapi_schema_manifest
+#
+# test/compat/test_openapi_compat.py compares the live manifest to this committed baseline.
```
GuillaumeDSM, line 9: this is the openapi.json breaking changes detector, an alternative would be to use a 3rd party lib for this but I prefer not to considering this is CI test only

Herklos (approved):
Good idea 👍

